PR #1, opened 2016-07-23
libplatsupport: Use 180 degree phase feedback for spi on exynos

agacek:
The comment in the code says to use a 180 degree phase feedback,
but the code was using no feedback clock. This commit changes the
code to match the comment (and fixes a bug we were experiencing on
the HACMS program).


agacek:
I changed these commits into issues so Data 61 can fix them as they see fit.

PR #6, opened 2016-12-07
Zync-7000 UART: Use status register to check if RX fifo is non-empty

agacek:
The QEMU emulator for Zync-7000 appears to have a bug makes it appear
like the UART timeout interrupt has fired even if the interrupt is
disabled. This causes the libplatsupport driver to make a read even
when no data is available (returning 0x00). This patch changes the
reading guard to check the status register directly rather than
checking interrupts. This works correctly under QEMU and should
continue to work on the actual hardware.


agacek:
Thanks. I also submitted [a patch to QEMU](http://lists.nongnu.org/archive/html/qemu-devel/2016-12/msg01057.html) which looks like it will be approved. So this (eventually) won't be a problem.

PR #7, opened 2017-01-13
libplatsupport: Fix tk1 uart rx fifo

agacek:


AdrianDanis:
Thanks @agacek, looks fine, should merge tonight after regression passes

PR #8, opened 2017-01-15
libplatsupport: Rewrite TK1 spi driver

agacek:
The previous TK1 spi driver had a series of issues which it made it
easier to replace rather than patch.


AdrianDanis:
Well I cannot comment on the functionality so I'll trust you that it's correct. The new code looks a lot nicer though

PR #11, opened 2017-06-07
Fixed warnings found by clang

mokshasoft:
~/seL4test/libs/libplatsupport/src/arch/arm/clock.c:45:22: warning: comparison of unsigned enum expression >= 0 is always true [-Wtautological-compare]
    ((void)((clk->id >= 0 && clk->id < NCLOCKS) || (__assert_fail("clk->id >= 0 && clk->id < NCLOCKS", "~/seL4test/libs/libplatsupport/src/arch/arm/clock.c", 45, __func__),0)));
             ~~~~~~~ ^  ~
~/seL4test/libs/libplatsupport/src/arch/arm/clock.c:70:12: warning: comparison of unsigned enum expression < 0 is always false [-Wtautological-compare]
    if (id < 0 || id >= NCLOCKS) {
        ~~ ^ ~
~/seL4test/libs/libplatsupport/src/arch/arm/clock.c:105:12: warning: comparison of unsigned enum expression < 0 is always false [-Wtautological-compare]
    if (id < 0 || id >= NCLOCKS) {
        ~~ ^ ~
~/seL4test/libs/libplatsupport/src/arch/arm/clock.c:116:12: warning: comparison of unsigned enum expression < 0 is always false [-Wtautological-compare]
    if (id < 0 || id >= NCLOCKS) {
        ~~ ^ ~
4 warnings generated.

### DIFF
--- a/libplatsupport/src/arch/arm/clock.c
+++ b/libplatsupport/src/arch/arm/clock.c
@@ -44,7 +44,7 @@ _fixed_clk_init(clk_t* clk)
 freq_t
 _default_clk_get_freq(clk_t* clk)
 {
-    assert(clk->id >= 0 && clk->id < NCLOCKS);
+    assert(clk->id < NCLOCKS);
     return ps_freq_default[clk->id];
 }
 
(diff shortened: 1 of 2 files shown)


mokshasoft:
make compile-all passes except for kzm_simulation_debug_xml_defconfig which also fails on my machine without this patch. All tests pass when running 'make simulate-x86_64' and when running the tests on a BeagleBone Black using config bbone_black_debug_xml_defconfig

PR #13, opened 2017-09-25
7.0.x dw

chrisguikema:
This is an update to the util_libs repository implementing Ethernet Drivers for the zynq7000, updates to the i.MX6 Ethernet Driver to make it more functional with the Ethdriver from the CAmkES-VM, and functionality for both Ethdrivers implementing promiscuous mode, which allows for multiple seL4Ethdriver connections.


chrisguikema:
You can remove the Add DARPA release statement. We need to keep it for our own records, though by now, you probably already know this.

AdrianDanis:
Hi @chrisguikema 
Thanks for this contribution. I will try to review this and get back to you this week.

libethdrivers/include/ethdrivers/raw.h
```diff
@@ -156,4 +156,11 @@ struct dma_buf_cookie {
     void* pbuf;
 };
 
+struct eth_plat_config {
```
AdrianDanis: Given this is a driver specific configuration struct this should not be in a generic header

libethdrivers/include/ethdrivers/raw.h
```diff
@@ -156,4 +156,11 @@ struct dma_buf_cookie {
...
+    void* buffer_addr;
+    uint8_t mac_addr[6];
+    uint8_t prom_mode;
+    ethif_driver_init init;
```
AdrianDanis: As far a I can tell this member is never referenced, so I cannot divine it's intended purpose

libethdrivers/src/plat/imx6/imx6.c
```diff
@@ -393,10 +395,17 @@ int ethif_imx6_init(struct eth_driver *eth_driver, ps_io_ops_t io_ops, void *con
     }
     eth_data->enet = enet;
 
-    /* Fetch and set the MAC address */
-    if (ocotp == NULL || ocotp_get_mac(ocotp, mac)) {
-        memcpy(mac, DEFAULT_MAC, 6);
+    if (plat_config->prom_mode) {
```
AdrianDanis: This is a bit of a breaking change, given that we have code that passes in `NULL` as the `config` ([as specified in the function description](https://github.com/seL4/libethdrivers/blob/master/include/ethdrivers/imx6.h#L25)). Perhaps the config should be made an optional parameter?

libethdrivers/src/plat/zynq7000/zynq7000.c
```diff
@@ -1,6 +1,5 @@
 /*
- * Copyright 2014, NICTA
```
AdrianDanis: Not sure why this copyright is dropped. There are large amounts of new implementation, but also large amounts of unchanged code and code with minor name/formatting changes

libethdrivers/include/ethdrivers/zynq7000.h
```diff
@@ -0,0 +1,29 @@
...
+
+#define ZYNQ7000_INTERRUPT_ETH0 54
+
+/**
+ * This function initialises the hardware and conforms to the ethif_driver_init
+ * type in raw.h
+ * @param[out] eth_driver   Ethernet driver structure to fill out
+ * @param[in] io_ops        A structure containing os specific data and
+ *                          functions.
+ * @param[in] config        Unused, should be NULL
```
AdrianDanis: this is true right now in this commit, but config becomes a required parameter later without this being updated

libethdrivers/include/ethdrivers/raw.h
chrisguikema: This change was ARM specific. There wasn't an arch_include folder, so we put it in here. What would your recommendation be?

libethdrivers/include/ethdrivers/raw.h
chrisguikema: This is used to make an platform independent driver. Instead of doing 

```#ifdef CONFIG_PLAT_IMX6
ethif_imx6_init
#elif CONFIG_PLAT_ZYNQ7000
ethif_zynq7000_init
#endif
```
So the Ethdriver component now has an platform init function call declared locally that sets the init member, so the function call is 
`    error = config->init(&eth_driver, ioops, (void *)config);`

libethdrivers/include/ethdrivers/zynq7000.h
chrisguikema: You're right. We will update the comment.

libethdrivers/src/plat/imx6/imx6.c
chrisguikema: Do you have an example using the i.MX6 driver anywhere? We only saw x86 examples. We could work together to figure out ways to improve this.

libethdrivers/src/plat/zynq7000/zynq7000.c
chrisguikema: You are right. We did a lot of commit squashing with these files. Feel free to reinsert it!

libethdrivers/include/ethdrivers/raw.h
chrisguikema: For reference, here is how we implemented an Ethdriver that functions with the zynq7000 and the imx6.

https://github.com/dornerworks/camkes-debug-tool/blob/master/components/Ethdriver/src/ethdriver.c

libethdrivers/src/plat/imx6/imx6.c
AdrianDanis: Okay so the example I'm thinking of is from our [Advanced Operating Systems course](http://www.cse.unsw.edu.au/~cs9242/current/), but it is pinned to an old version of the code anyway so doesn't matter. I would be happy with updating the comment in the header and then having an `assert(config)`

libethdrivers/include/ethdrivers/raw.h
AdrianDanis: My objection here is that the `init` function here is not a configuration to the driver, it's a configuration in your component. I would suggest having your own config struct in your component like

```c
struct ethdriver_config {
    eth_plat_config driver_config;
    ethif_driver_init driver_init;
};
```

libethdrivers/include/ethdrivers/raw.h
AdrianDanis: The current structure of this library leaves much to be desired. I do not want to make you restructure it, so for now maybe call it `arm_eth_plat_config`, and leave a comment saying it's for the imx6/zynq and should be moved somewhere else at some point?

libethdrivers/src/plat/imx6/imx6.c
chrisguikema: Alright. We will update the header command and add in the assert value.

libethdrivers/include/ethdrivers/raw.h
chrisguikema: Understood. We will remove the component specific `init`  from the driver structure.

libethdrivers/include/ethdrivers/raw.h
chrisguikema: Okay.

chrisguikema:
I'll update the Pull Request with your suggestions later today.

chrisguikema:
Any update?

AdrianDanis:
Sorry @chrisguikema, I failed to manually check this PR and github does not send notifications when commits in a PR are changed. Will take a look now.

libethdrivers/Kbuild
```diff
@@ -13,4 +13,4 @@
 libs-$(CONFIG_LIB_ETHIF) += libethdrivers
 libethdrivers-${CONFIG_LIB_LWIP} += liblwip
 libethdrivers-${CONFIG_LIB_PICOTCP} += libpicotcp
-libethdrivers: ${libethdrivers-y} libutils libsel4 $(libc) libplatsupport common
+libethdrivers: ${libethdrivers-y} libutils libsel4 libsel4utils $(libc) libplatsupport common
```
AdrianDanis: Just a minor question of why libsel4utils needed to be added as a dependency? I don't see any usage of things from it in this commit, were there some pre-existing usages that were broken prior to this being added?

AdrianDanis:
Aside from the one question I had your changes look fine. Sorry again for the delay. I will pull it in and rebase and remove the DARPA release statement. Hopefully appears next week, depending on if we fix our builds by then or not.

libethdrivers/Kbuild
chrisguikema: I don't remember why this was added. I just removed the dependency and the library builds for both the zynq7000 and the i.MX6 without any issues. This can be removed.

AdrianDanis:
Closing this now as a rebased version has been merged.

PR #14, opened 2018-06-08
Fixed memory initialization for preallocated RX ring buffer

fabriziobertocci:
RX Ring buffer (rx_cookie) was corrupt and uninitialized causing the system to throw an invalid memory access violation when a new packet was received from the iMX6 ethernet controller.

Fix change the definition of the rx_cookie to be a void * where the actual buffer and meaning is delegated to the file that interface with the stack (in my case I'm using lwip.c).

See commit comments for additional info


AdrianDanis:
@fabriziobertocci these changes look good. As you can tell we're not currently using the imx6 ethdriver internally, hence the bitrot, so thanks for fixing that.

As I said in the other PR, still waiting to confirm CLA

PR #15, opened 2018-06-21
Small change: removed unused variable...

fabriziobertocci:
... to avoid compiler's warnings


AdrianDanis:
@fabriziobertocci Looks like the [sglib](https://github.com/stefanct/sglib) got updated (some time ago) so we should probably just wholesale import the new version.

PR #20, opened 2019-01-16
Use DMA Buffer Structure for allocating RX Buffers.

chrisguikema:
This method is used in the zynq7000 ethernet driver
instead of a void pointer. While testing on the lastest
release, DornerWorks noticed that the imx6 would pass a
NULL pointer to the Ethdriver component, which was not
behavior noticed in previous releases.


chrisguikema:
Just added a second commit so libethdrivers compiles with zynq7000

PR #23, opened 2019-07-12
Generic timer

pingerino:
Update generic timer to work for aarch64 as well as aarch32.

Add a generic timer ltimer implementation which can be included for platforms.

### DIFF
--- a/libplatsupport/CMakeLists.txt
+++ b/libplatsupport/CMakeLists.txt
@@ -71,4 +71,7 @@ target_include_directories(
 if(NOT "${KernelArmMach}" STREQUAL "")
     target_include_directories(platsupport PUBLIC mach_include/${KernelArmMach})
 endif()
+if("${KernelArch}" STREQUAL "arm")
+    target_include_directories(platsupport PUBLIC sel4_arch_include/${KernelSel4Arch})
+endif()
 target_link_libraries(platsupport Configuration muslc utils)
--- a/libplatsupport/CMakeLists.txt
+++ b/libplatsupport/CMakeLists.txt
@@ -69,7 +69,6 @@ if(KernelPlatImx6 OR KernelPlatformImx7Sabre OR KernelPlatformKZM OR KernelPlatf
     endif()
 endif()
 
-
 list(SORT deps)
 
 add_config_library(platsupport "${configure_string}")
--- a/libplatsupport/CMakeLists.txt
+++ b/libplatsupport/CMakeLists.txt
@@ -37,10 +37,22 @@ file(
         deps
         src/mach/${KernelArmMach}/*.c
         src/plat/${KernelPlatform}/*.c
-        src/arch/${KernelArch}/*.c
         src/*.c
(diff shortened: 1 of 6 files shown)


libplatsupport/src/arch/arm/generic_ltimer.c
```diff
@@ -0,0 +1,183 @@
...
+        error = ENXIO;
+    } else {
+        generic_ltimer->ops = ops;
+        generic_timer_enable();
+    }
+
+    /* if there was an error, clean up */
+    if (error) {
+        ZF_LOGE("Failed to initialise generic timer");
+        destroy(ltimer);
```
WellMcGarnicle: Should this be `ltimer->data`?

libplatsupport/src/arch/arm/generic_ltimer.c
```diff
@@ -0,0 +1,183 @@
+/*
+ * Copyright 2028, Data61
```
WellMcGarnicle: Typo in the year ( I think its still 2019 )

libplatsupport/CMakeLists.txt
```diff
@@ -42,6 +42,18 @@ file(
         src/plat/${KernelPlatform}/acpi/*.c
 )
 
+if(${KernelArch} STREQUAL "arm")
```
kent-mcleod: Why add these in addition to the file(GLOB ..) call above?

pingerino: because I should have also deleted the glob here. Updated. I don't want to compile in the generic_ltimer if its not selected for the platform specifically. On my branch I have 

```
if(KernelPlatformSDM845)
    list(APPEND deps src/arch/arm/generic_ltimer.c)
endif()
```

kent-mcleod: FYI the commit message in 41f0b2a8af9393fec655a05cd7d9a5d1caf01504 says you always compile the generic_timer.

kent-mcleod: Ah, one refers to generic_timer.c and the other to generic_ltimer.c.  I understand now

kent-mcleod: Do you want to add `  list(APPEND deps src/arch/arm/generic_timer.c) ` here so that it is the same as it was before?

libplatsupport/src/arch/arm/generic_ltimer.c
```diff
@@ -0,0 +1,183 @@
...
+        error = ENXIO;
+    } else {
+        generic_ltimer->ops = ops;
+        generic_timer_enable();
+    }
+
+    /* if there was an error, clean up */
+    if (error) {
+        ZF_LOGE("Failed to initialise generic timer");
+        destroy(data);
```
WellMcGarnicle: Doesn't look like `data` is declared anywhere?

libplatsupport/CMakeLists.txt
pingerino: Done

PR #25, opened 2019-08-15
Zcu102

chrisguikema:
These commits are necessary for serial server and time server support on the zynqmp.

### DIFF
--- a/libplatsupport/mach_include/zynq/platsupport/mach/timer.h
+++ b/libplatsupport/mach_include/zynq/platsupport/mach/timer.h
@@ -174,4 +174,4 @@ void ttc_freerun(ttc_t *tcc);
 int ttc_handle_irq(ttc_t *ttc);
 uint64_t ttc_get_time(ttc_t *ttc);
 /* convert from a ticks value to ns for a configured ttc */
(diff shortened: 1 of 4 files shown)


chrisguikema:
@nomadeel Now that the global-components commits have been merged in, these commits are necessary for proper zynqmp support.

On the timer side, the free-running clock will only run for ~38 seconds, which is how long the timer takes to overflow. 

The serial server will print just fine, but the interrupt handler wasn't implemented, so when getting a character, the interrupt would never actually be cleared in hardware, which would continually trigger the serial server to get a character, even if one wasn't in the fifo.

nomadeel:
Thanks for the patches! I forgot to do this for the last PR in global-components, but could you modify some of the commit messages so that:
- the subject is less than 50 characters, and
- there's a small message in the body which describes what the commit does?

This is mainly to have our git style checker happy with these commits. 

Also, could you rebase against master or would you like me to do it? This is so that Github can detect that the patches are released into master and this PR will be closed nicely when that happens.

chrisguikema:
Alright, I updated the commit messages per the Data61 standard and rebased against the current master. Everything should be good to go!

nomadeel:
Sweet, I'll pull in your patches now.

PR #31, opened 2019-11-06
changes for generic imx8m

ridale:
Change set to enable generic imx8m device, tested on 
aarch32 - imx8mm and imx8mq
aarch64 - imx8mm and imx8mq


ridale:
Will not be merged due to internal refactor

PR #39, opened 2020-09-09
Return allocated IRQ from FDT

xurtis:
When allocating an IRQ from the FDT the IRQ that was allocated should be passed back to the caller in addition to the IRQ handler ID.

For `rockpro64`, the `ltimer` implementation for its second timer should use the IRQ number one greater that the first timer's IRQ number (rather than the first timer's IRQ handler ID).

Fixes sel4/sel4test#25.

### DIFF
--- a/libplatsupport/plat_include/rockpro64/platsupport/plat/timer.h
+++ b/libplatsupport/plat_include/rockpro64/platsupport/plat/timer.h
@@ -51,6 +51,7 @@ typedef struct {
     volatile struct rk_map *hw;
     void *rk_map_base;
     pmem_region_t pmem;
+    ps_irq_t irq;
     irq_id_t irq_id;
 } rk_t;
 
--- a/libplatsupport/src/plat/rockpro64/timer.c
+++ b/libplatsupport/src/plat/rockpro64/timer.c
@@ -15,6 +15,7 @@
 #include <errno.h>
 #include <utils/util.h>
 #include <platsupport/timer.h>
+#include <platsupport/fdt.h>
 #include <platsupport/plat/timer.h>
 #include <utils/frequency.h>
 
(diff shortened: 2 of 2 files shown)


xurtis:
This passes all tests internally including tests with SMP enabled on rockpro64

xurtis:
`IPC1001` seems to get stuck with all 6 cores enabled still which is an improvement over sel4/sel4#184

yyshen:
how long did you wait? it might take a while.

xurtis:
Well, I looked and it's now up to IPC1002 so it seems to be taking an astoundingly long time.

yyshen:
ok I expected it will take a while. similar story on hikey when all 8 cores are enabled.

xurtis:
`MULTICORE0002` and `MULTICORE0005` fail on the 6-core config for `rockpro64`

yyshen:
> `MULTICORE0002` and `MULTICORE0005` fail on the 6-core config for `rockpro64`

some logs?

xurtis:
If I run the multicore tests on their own they pass without issue

yyshen:
hmm, how about IPC and Multicore tests together?

xurtis:
I reran all of the tests and they all passed this time so :shrug:

kent-mcleod:
`ps_fdt_index_register_irq` is intended to have a simple interface to handle the common case of registering a callback handler on an interrupt at a certain index of the device tree.  In cases where you need more complicated behavior, which appears to be the case here, then `ps_fdt_walk_irqs` should be used instead which does provide the `ps_irq_t` and will allow you to handle the rockpro64 behavior.

xurtis:
@kent-mcleod updated the changes per your recommendation.

kent-mcleod (approved):

xurtis:
@ssrg-bamboo `test`

ssrg-bamboo:
Hello, I'm a bot! I'll bring this PR into Trustworthy Systems and run some tests

ssrg-bamboo:
All the tests we ran have passed! Nice job!

nomadeel (approved):

PR #73, opened 2021-03-12
use const qualifier for libcpio and libelf wherever possible

axel-h:
CPIO is handling fixed data, so it should use the `const` qualifier wherever possible. This allows passing it const data and use the type checking features from the compiler to find invalid usages. 

Note that this change will break the ElfLoader (at least in strict mode), as it does not use const either. There is a commit in https://github.com/seL4/seL4_tools/pull/54 that fixed this.

### DIFF
--- a/libcpio/include/cpio/cpio.h
+++ b/libcpio/include/cpio/cpio.h
@@ -49,7 +49,7 @@ struct cpio_info {
  * @return             The location of the file in memory; NULL if the index
  *                     exceeds the number of files in the CPIO archive.
  */
-void *cpio_get_entry(void *archive, unsigned long len, int index, const char **name, unsigned long *size);
+const void *cpio_get_entry(const void *archive, unsigned long len, int index, const char **name, unsigned long *size);
 
 /**
  * Retrieve file information from a provided file name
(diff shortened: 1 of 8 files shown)


libcpio/include/cpio/cpio.h
```diff
@@ -49,7 +49,12 @@ struct cpio_info {
  * @return             The location of the file in memory; NULL if the index
  *                     exceeds the number of files in the CPIO archive.
  */
-void *cpio_get_entry(void *archive, unsigned long len, int index, const char **name, unsigned long *size);
+void *cpio_get_entry(
```
kent-mcleod: Are you able to open an issue somewhere about proposing a change for adopting this different style and recording it in the style guide? The current guide allows for function arguments to continue on the same line until it reaches the line length I believe.  I believe that we generally try and avoid any stylistic changes that haven't been generated by the style tool to prevent stylistic arguments from sidetracking PRs as everyone has different opinions on what the better style is.  For me, I like more arguments on one line :)

axel-h: I can do this. For a few params, having it on the same line makes sense, especially if the are closely related (e.g. buffer/length). The rationale for having the parameter one on each line it that it allows better diffs/patch/merge workflow, as you (or the tools) can see the actual change better.

pingerino: Agree with @kent-mcleod . Please argue for changing the style on an issue before doing it. We have a style guide already, we shouldn't go an restyle code without buyin, as there are trade-offs that need to be considered. 

Additionally, I believe I turned this *off* in our astyle currently :)

oliver-wm:
Can this be changed to remove the style.

axel-h:
ok, will remove the style change

oliver-wm (approved):

axel-h:
@oliver-wm : I still need to address the comments, please do not merge yet.

axel-h:
I've addresses the comments, should be ready to merge now together with https://github.com/seL4/seL4_libs/pull/29

PR #82, opened 2021-04-09
libsatadrivers: add sata support

nomadeel:
This commit combines a number of smaller commits in order add sata
support. This commit does the following:

    * Add libsatadrivers from previous project
    * Convert libsatadrivers to cmake
    * Convert libsatadrivers to DPRINTF
    * Fix io ops memory issue in libsatadrivers
    * Change IDE command and control io port registers
    * Fix LBA 48 bit PIO implementation
    * Fix buffer overflow in PIO write
    * Refactor satadrivers library
    * Add AHCI compatibility to libsatadrivers
    * Add more ahci error handling and reset functionality
    * Make helper functions static
    * Allow partition data to be sent back to caller
    * Add null pointer check to common functions
    * Fix compile time warnings
    * dprintf -> zf_log
    * Print every ahci register
    * Replace wait on FR bit with a delay

CCDC-GVSC DISTRIBUTION A.  Approved for public release; distribution
unlimited. OPSEC#4481.

Signed-off-by: Damon Lee <Damon.Lee@data61.csiro.au>


nomadeel:
This PR is based off this commit from Dornerworks' fork 716c01545e7fa3ba4f9af7950e9be792becea671.

xurtis:
Signed-off-by implies that you hold the copyright to the commit...

xurtis:
Also looks like all the licenses need to be updated

xurtis:
why is all of this a single commit?

libsatadrivers/include/satadrivers/raw.h
```diff
@@ -0,0 +1,100 @@
...
+#pragma once
+
+#include <stdint.h>
+#include <stdlib.h>
+#include <platsupport/io.h>
+#include <satadrivers/virtio/virtio_blk.h>
+
+struct disk_driver;
+
+#define VIRTIO_BLK_XFER_FAILED   -1
```
axel-h, line 19: put this in brackets to make it an atom

libsatadrivers/include/satadrivers/ide.h
```diff
@@ -0,0 +1,148 @@
...
+#define LBA_48_MODE 2
+#define LBA_28_MODE 1
+#define LBA_CHS_MODE 0
+
+#define NUM_IDE_CHANNELS  2
+
+#define NUM_SECT_PER_TRACK 63
+#define NUM_LBA_BYTES 6
+
+#define IDENT_LBA48_SUPPORT_BIT (1 << 26)
```
axel-h, line 144: use "1u"

libsatadrivers/include/satadrivers/virtio/virtio_blk.h
```diff
@@ -0,0 +1,126 @@
+/* This header is BSD licensed so anyone can use the definitions to implement
```
axel-h, line 1: consider adding an SPDI tag

lsf37:
> Signed-off-by implies that you hold the copyright to the commit...

Not necessarily, it certifies that:
```
(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```
So if you are sure that the commits were previously provided under the right licenses or if you have the right to release them, signing off on them is fine.

libsatadrivers/LICENSE_BSD2.txt
```diff
@@ -0,0 +1,30 @@
+Files described as being under the "BSD 2-Clause" license fall under the
```
lsf37, line 1: This file should not be included, instead there should be SPDX tags in the sources, and the corresponding license files under `LICENSES/`

libsatadrivers/LICENSE_GPLv2.txt
```diff
@@ -0,0 +1,295 @@
+
+Files described as being under the "GNU General Public License version 2"
```
lsf37, line 2: (Same as BSD license, should now be done with SPDX)

libsatadrivers/include/satadrivers/virtio/virtio_blk.h
lsf37: Yes, please. Here the tag should be added, not replacing the existing header (only replace for our own headers)

lsf37 (changes requested):
The license headers have to be fixed up first. If you need help with that, let me know.

lsf37:
The line

> CCDC-GVSC DISTRIBUTION A. Approved for public release; distribution unlimited. OPSEC#4481.

sound like it was originally part of one or more headers. If it was, it should stay there.

libsatadrivers/include/satadrivers/ahci.h
```diff
@@ -0,0 +1,117 @@
...
+#define SECTORS_TO_BYTES(num) ((num<<9)-1) // 512 bytes per sector
+
+#define DATA_BLK_SIZE      4095   // 4K bytes (this value should always be set to 1 less than the actual value)
+#define BYTES_PER_PRDT     4096
+#define SECTORS_PER_PRDT   8
+
+#define TIMEOUT_10S 10000
+#define TIMEOUT_1S  1000
+
+#define PxSERR_RESERVED_BITS(x) (((x) >= 2 && (x) <= 7) || ((x) >= 12 && (x) <= 15))
```
axel-h, line 86: The macro argument is used several times, that is potentially dangerous. Make this static inline function or make this macro create a "local" variable first using `typeof(x) _x_ = (x); ...`

libsatadrivers/include/satadrivers/ide.h
```diff
@@ -0,0 +1,148 @@
...
+
+// IDE Control Registers
+#define ATA_REG_CONTROL    0x02
+#define ATA_REG_ALTSTATUS  0x02
+
+#define PRIMARY_CMD_BASE   0x01F0
+#define PRIMARY_CTRL_BASE  0x03F4
+
+#define DELAY_CYCLES       10000000
+#define SATA_DELAY(delay) do { for(int x = 0; x < (delay); x++); } while(0);
```
axel-h, line 108: looks like a hack. Make this function can then add something in the loop that ensures this does not get optimized out.

libsatadrivers/src/ahci.c
```diff
@@ -0,0 +1,956 @@
...
+#include <satadrivers/ahci.h>
+#include <unistd.h>
+#include <string.h>
+#include <sel4/sel4.h>
+#include <platsupport/delay.h>
+
+#include <satadrivers/ahci_types.h>
+#include <utils/zf_log.h>
+
+#define GET_LOWER_32BITS(val) (uint32_t)((uint64_t)(val) & 0xFFFFFFFFLL)
```
axel-h, line 25: why not simply `((uint32_t)(val))`

libsatadrivers/src/ahci.c
```diff
@@ -0,0 +1,956 @@
...
+#define GET_UPPER_32BITS(val) (uint32_t)(((uint64_t)(val) & 0xFFFFFFFF00000000LL) >> 32)
```
axel-h, line 26: why not simply `((uint32_t)((val) >> 32))`

axel-h:
Is there some example also that show how to use it?

nomadeel:
Gah, why doesn't GitHub have threading.

From @xurtis:
> why is all of this a single commit?

This is how it was release onto GitHub by people from Dornerworks. It's a lot of work to split up the giant commit into smaller commits and there isn't much value to do so, I think.

From @lsf37:
> The line
> 
> > CCDC-GVSC DISTRIBUTION A. Approved for public release; distribution unlimited. OPSEC#4481.
> 
> sound like it was originally part of one or more headers. If it was, it should stay there.

It's not part of the headers but was inside the commit message, so I'll leave it inside the commit message.

From @axel-h: 
> Is there some example also that show how to use it?

Yeah, there is. The Sataserver CAmkES component in this [patch](https://github.com/seL4/camkes-vm/pull/8) as well as the emulated VirtIO block driver in this [patch](https://github.com/seL4/seL4_projects_libs/pull/17) use the driver. This [sample CAmkES VM app](https://github.com/seL4/camkes-vm-examples/pull/4) integrates all of this to virtualise a partition on a hard drive to the guest Linux VM.

libsatadrivers/LICENSE_BSD2.txt
nomadeel: Right, will do. Also have to fix up all files' licenses as well.

libsatadrivers/include/satadrivers/ahci.h
nomadeel: I'll make it a static inline function.

libsatadrivers/include/satadrivers/ide.h
nomadeel: Will something like this work?

E.g.
```
int i = 0;
while (i < delay) {
    i++;
}
```

libsatadrivers/include/satadrivers/ide.h
nomadeel: Yep,

libsatadrivers/include/satadrivers/raw.h
nomadeel: Yep.

libsatadrivers/include/satadrivers/virtio/virtio_blk.h
nomadeel: Yep, also have to fix other files as well.

libsatadrivers/src/ahci.c
nomadeel: Added assurance I suppose. Not too sure, I didn't write it.

libsatadrivers/src/ahci.c
nomadeel: Same as the `LOWER_32BITS`, added assurance.

axel-h: C rules are pretty clear how things works. This looks a bit like mystification unless there is an explanation why all this mystification this is needed.

libsatadrivers/include/satadrivers/ide.h
axel-h: That will not really change a thing, as the optimized will still concludes this does nothing. A common thing is to loop over NOP:
```
void busy_loop(unsigned int n) {
    while (n-- > 0) {
        __asm__ volatile("nop");
    }
}
```
But that assumes there is a "nop" instruction. There is also a way to really loop over nothing by adding an empty  assembly block that just clains it need the loop variables. So far, no optimized dares to remove this.
```
void busy_loop(unsigned int n) {
    while (n-- > 0) {
        __asm__ volatile("" : "+g" (i) : :);
    }
}
```
But it still leaves the open question, how long this will actually waits. Instead of `SATA_DELAY()`, better use something `usleep()` that at least give some indication how long this is supposed to wait.

libsatadrivers/src/ahci.c
nomadeel: Right, ok, I'm happy to just follow the suggestion then.

libsatadrivers/include/satadrivers/ide.h
oliver-wm: Out of interest, do you know if there is an attribute that tells the compiler not to optimize?

oliver-wm:
Yes I agree, it's too much work to split this into small commits. This should have happened on the original PR.

libsatadrivers/include/satadrivers/ide.h
nomadeel: I'm assuming that it just waits for a very small moment or something. The intended counter is 10000000, which on a 3.5GHz core, is about 2ms. Given that it only happens in the init phase, I don't think this warrants a proper `usleep()` implementation.

axel-h: * @nomadeel : We could still call this function `busy_loop_us_sleep()`. The the code using this then would call `busy_loop_us_sleep(2000)` to state that it wants to wait 2ms. The implementation of the function can then loop `5000*n` and say, that this is a crude hack where test have shown that passing 5.000.000 loops on 3.5GHz core are about 1 ms.
* @oliver-wm: see https://gcc.gnu.org/onlinedocs/gcc-10.3.0/gcc/Function-Specific-Option-Pragmas.html#Function-Specific-Option-Pragmas and https://gcc.gnu.org/onlinedocs/gcc-10.3.0/gcc/Common-Function-Attributes.html#Common-Function-Attributes#index-optimize-function-attribute, there is __attribute__((optimize("O0")))` or 
```
#pragma GCC push_options
#pragma GCC optimize ("O0")
... code ...
#pragma GCC pop_options
```

libsatadrivers/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+cmake_minimum_required(VERSION 3.7.2)
+
+project(libsatadrivers C)
+
+set(configure_string "")
+
+config_option(LibSatadrivers LIB_SATA "Support library for x86 SATA" DEFAULT OFF)
+
+config_option(
+    LibSatadriversAHCIEnable
```
kent-mcleod, line 16: A more CMake-friendly way of implementing this option would be to create two libraries: satadrivers-ahci and satadrivers-ide.  Then the application can specify which library to link, but the mode no longer has to be a global setting which would disallow one component from having an achi based driver and another having ide.

libsatadrivers/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+#
+# SPDX-License-Identifier: BSD-2-Clause
+#
+cmake_minimum_required(VERSION 3.7.2)
+
+project(libsatadrivers C)
+
+set(configure_string "")
+
+config_option(LibSatadrivers LIB_SATA "Support library for x86 SATA" DEFAULT OFF)
```
kent-mcleod, line 13: What is this config used for? Is it still required?

nomadeel: Nope, that's not used. We can remove it.

libsatadrivers/CMakeLists.txt
nomadeel: I think I could refactor parts of the library to allow both the AHCI and IDE drivers to co-exist. The users can then just pick whichever one to use depending on their needs.

lsf37 (approved):
This now looks good from my side.

PR #83, opened 2021-04-10
i.MX6 NIC driver cleanup, make it work on QEMU/sabre

axel-h:
* i.MX6 NIC driver code cleanup
* make it work on QEMU/sabre (ie run on bare metal without U-Boot that sets up the clocks)
* set a new baseline for further (pending) pull request


axel-h:
What would be the way to get this merged? The first commit is needed to make things compile again. Then there is a lot of refactoring to make the code more readable, scalable and clarify what is going on. The last two commits change things a bit.
* one commit removes the checksum from the ethernet packet on the rationale, that nobody really cares about it anyway - unless there is a know use case where the whole packet must be forwarded. But then a configuration interface would be needed. 
* the other commit add a way to configure the driver in a better ways. We are using this in our CAmkES driver. This will also be needed for the next PR I am about to create,  that will bring in the i.MX6sx support where 2 ethernet ports exists that this driver can support then. 

I can remove the last 2 commits from the current PR (and add them in to the next PR then) if this is the only real blocker here. That would still keep the improvement driver working in QEMU/sabre, because this is mainly due to the clock setup that is done in the first few commits of this PR.

oliver-wm:
> * one commit removes the checksum from the ethernet packet on the rationale, that nobody really cares about it anyway - unless there is a know use case where the whole packet must be forwarded. But then a configuration interface would be needed.

I think it would be better to keep it but be able to configure it somehow.

> * I can remove the last 2 commits from the current PR (and add them in to the next PR then) if this is the only real blocker here.

Might be a good idea to keep all the major changes in one pr and all the cosmetic / style changes in another (or smaller) pr(s).

axel-h:
I will make this configurable then via the mechanism the last commit in the current PR adds.

axel-h:
I've merged the last two commits now. The default behavior remains unchanged, but you can provide a config also to set things up specifically.

libethdrivers/src/helpers.c
```diff
@@ -1,33 +1,41 @@
...
 #include <ethdrivers/helpers.h>
 
-dma_addr_t
-dma_alloc_pin(ps_dma_man_t *dma_man, size_t size, int cached, int alignment)
+dma_addr_t dma_alloc_pin(ps_dma_man_t *dma_man, size_t size, int cached, int alignment)
 {
     void *virt = ps_dma_alloc(dma_man, size, alignment, cached, PS_MEM_NORMAL);
     if (!virt) {
-        return (dma_addr_t) {0, 0};
+        LOG_ERROR("ps_dma_alloc() failed for size=%zu, alignment=%d",
```
pingerino, line 14: These should use `ZF_LOGE` in any new code to be consistent. The only LOG_ERROR macros are from a time when every library invented it's own logging macros and are deprecated.

libethdrivers/src/plat/imx6/uboot/phy.c
```diff
@@ -51,69 +45,81 @@
...
-					adv);
-
-			if (err < 0)
-				return err;
-			changed = 1;
-		}
-	}
-
-	return changed;
+    u32 advertise;
```
pingerino, line 48: I think this code was pulled from uboot and shouldn't be restyled

libethdrivers/src/helpers.c
axel-h: I will change this

libethdrivers/src/plat/imx6/uboot/phy.c
axel-h: It's hard to argue with the automatic style checker, it will always complain about this file for each commit that changes one line. That is why I made one commit that fixes the style now. I have some doubts this is every really synced with U-Boot again, as manche thing see completely different by now in the head version.

axel-h:
https://github.com/seL4/util_libs/pull/87 holds the first commit form this now, as this fixes a build breaker.

axel-h:
I've changed the logging everywhere now to no longer use the deprecated functions.

libethdrivers/src/plat/imx6/uboot/phy.c
lsf37: You can add the file to `.stylefilter`, which if it doesn't exist yet should be at the repo root. This will cause the checker to skip the checks for this file. If it has diverged a lot already that might not be worth it any more, but if not it's an option.

axel-h: It's up to you to demand this. But I'd prefer no to rework this patch again, since the code has been reworked a lot already and diverges from U-Boot. There are also follow-up patches for the Nitrogen board that I have to rework then.

oliver-wm: Does it really matter that much if there are slight style changes?

libplatsupport/src/plat/imx6/clock.c
```diff
@@ -322,60 +329,110 @@ static struct clock arm_clk = { CLK_OPS(ARM, arm, NULL) };
...
+         * 0x3 source is CLK1_N/CLK1_P XOR CLK2_N/CLK2_P
+         */
+        unsigned int src = PLL_GET_BYPASS_SRC(v);
+        switch (src) {
+        case 0:
+            return 24 * MHZ;
+        default:
+            break;
+        }
+        ZF_LOGE("can determine frequency for bypass sources %u", src);
```
nomadeel, line 360: "can determine" -> "can't determine"?

libplatsupport/src/plat/imx6/clock.c
```diff
@@ -576,11 +637,16 @@ static struct clock mmdc_ch0_clk = { CLK_OPS(MMDC_CH0, mmdc_ch0, NULL) };
 
 static freq_t _ahb_get_freq(clk_t *clk)
 {
+    /* ToDo: read divider from CBCDR.AHB_PODF */
```
nomadeel, line 642: Is this TODO intentional and not something that was forgotten?

libplatsupport/src/plat/imx6/clock.c
```diff
@@ -576,11 +637,16 @@ static struct clock mmdc_ch0_clk = { CLK_OPS(MMDC_CH0, mmdc_ch0, NULL) };
...
     return clk_get_freq(clk->parent) / 4;
 }
 
 static freq_t _ahb_set_freq(clk_t *clk, freq_t hz)
 {
+    /* ToDo: we assume the default value CBCDR.AHB_PODF = b011 has not been
```
nomadeel, line 648: Same here: Is this TODO intentional and not something that was forgotten?

libplatsupport/src/plat/imx6/clock.c
```diff
@@ -71,15 +71,16 @@
...
-#define CLKO2_SRC_MASK      (0x1FU << 16)
-#define CLKO2_ENABLE        (1U << 24)
-#define CLKO_SEL            (1U << 8)
+/* CCM Clock Output Source (CCM_CCOSR) */
+#define CLKO1_SRC_AHB       0xBU
+#define CLKO1_SRC_IPG       0xCU
+#define CLKO1_SRC_MASK      0xFU /* bit 0-3 */
+#define CLKO1_ENABLE        BIT(7)
+#define CLKO_SEL            BIT(8) /* select CCM_CLKO1 or CCM_CLKO2 */
+#define CLKO2_SRC_MMDC_CH0  (0U << 16)
```
nomadeel, line 80: Can you confirm that this should be `(0U << 16)` and not the previous `(0U << 21)`?

libethdrivers/src/plat/imx6/imx6.c
```diff
@@ -348,32 +325,38 @@ static void complete_rx(imx6_eth_driver_t *dev)
...
-        assert(enet);
-        if (!enet_rx_enabled(enet)) {
-            enet_rx_enable(enet);
+        void *cookie = ring->cookies[head];
+        /* Go to next buffer, handle roll-over. */
+        if (++head == ring->cnt) {
+            head = 0;
         }
+        ring->head = head;
+        /* if rx and tx happens in parallel, this can have a race condition */
```
nomadeel, line 365: I don't see how the race condition can happen when the ring data structures are independent of each other. Am I missing something here?

nomadeel:
Sorry it took so long to review.

libethdrivers/src/plat/imx6/imx6.c
axel-h: I has hoping for some discussion here of there could be an issue or not. Basically, the queue could be use by two entities, a producer and a consumer. So the question is, how thread-safe it is or if the access if serialized somewhere else. The comment that that rx/tx happens in parallel is misleading, it's actually the producer and consumer running in parallel.

libplatsupport/src/plat/imx6/clock.c
axel-h: This is intentional in this comment I've just refactored the code and added comments to clarify the assumption made here.

libplatsupport/src/plat/imx6/clock.c
axel-h: This is intentional in this comment I've just refactored the code and added comments to clarify the assumption made here. It is something that should be done in the future if this is needed.

libethdrivers/src/plat/imx6/imx6.c
```diff
@@ -307,7 +317,7 @@ static int setup_desc_ring(imx6_eth_driver_t *dev, ring_ctx_t *ring)
                   sizeof(void *),
                   (void **) &ring->cookies);
     if (ret) {
-        LOG_ERROR("Failed to malloc, code %d", ret);
+        ZF_LOGE("Failed to malloc, code %d", ret);
```
axel-h, line 320: Note to self: this should have been fixed in an earlier cleanup commit already, there should not be an change in this commit for the logging.

libethdrivers/src/plat/imx6/uboot/phy.c
pingerino: Given the code is from U-boot it's best to keep it as close to original as possible to pull in updates if needed. Ideally we'd check U-boot mainstream to see if these patches already exist there.

oliver-wm (approved):

libethdrivers/src/plat/imx6/imx6.c
nomadeel: Right. This isn't thread-safe at all and access to this isn't serialized at all. But we've been getting along by using single-threaded environments and making sure that access is serialized.

nomadeel: In fact, all libraries in util_libs relies on this assumption.

libethdrivers/src/plat/imx6/enet.c
```diff
@@ -416,12 +416,14 @@ int enet_mdio_write(struct enet *enet, uint16_t phy, uint16_t reg, uint16_t data
...
  *** ENET driver ***
  *******************/
+
 void enet_rx_enable(struct enet *enet)
 {
     enet_get_regs(enet)->rdar = RDAR_RDAR;
 }
 
-int enet_rx_enabled(struct enet *enet)
+int enet_rx_enabled(
```
oliver-wm, line 425: Does this need to be reformatted?

libethdrivers/src/plat/imx6/enet.c
```diff
@@ -609,23 +621,25 @@ enet_init(struct desc_data desc_data, ps_io_ops_t *io_ops)
...
-    return ret;
+    return enet;
 }
 
 /****************************
  *** Debug and statistics ***
  ****************************/
 
-static void dump_regs(uint32_t *start, int size)
+static void dump_regs(
```
oliver-wm, line 640: reformat

oliver-wm (approved):

libethdrivers/src/plat/imx6/enet.c
axel-h: yes

libethdrivers/src/plat/imx6/imx6.c
axel-h: I'll improve the comment to state this

libplatsupport/src/plat/imx6/clock.c
axel-h: fixed

axel-h: fixed

libplatsupport/src/plat/imx6/clock.c
axel-h: Yes. The TRM says that bit 16-20 is `CLKO2_SEL` and the value `mmdc_ch0_clk_root` is b00000. Bit 21-23 is `CLKO2_DIV`. The function `_clko_init()` will set this value:
```
            v &= ~CLKO2_SRC_MASK;
            v |= CLKO2_SRC_MMDC_CH0;
```
The old value seem wrong. But it has worked because it is all zeros anyway. Anyway, I've made a dedicated commit "libplatsupport/imx6: fix CLKO2_SRC_MMDC_CH0" for this now.

libethdrivers/src/plat/imx6/imx6.c
axel-h: fixed

libethdrivers/src/plat/imx6/enet.c
axel-h: fixed

libethdrivers/src/plat/imx6/enet.c
axel-h: fixed

libethdrivers/src/plat/imx6/imx6.c
axel-h: done

libethdrivers/src/plat/imx6/imx6.c
```diff
@@ -590,11 +591,23 @@ static int init_device(imx6_eth_driver_t *dev)
...
+    enet_set_speed(
+        dev->enet,
+        dev->phy->speed,
+        (dev->phy->duplex == DUPLEX_FULL) ? 1 : 0);
+
+    ZF_LOGI("Link speed: %d Mbps, %s-duplex",
+            dev->phy->speed,
+            (dev->phy->duplex == DUPLEX_FULL) ? "full" : "half");
+
+    udelay(100000); /* why? */
```
kent-mcleod: My guess is that the delay is probably to give time for the link speed setting to propagate.  There's probably a register to poll for completion of the setting, but I wouldn't know what it would be.

kent-mcleod (approved):

libplatsupport/src/plat/imx6/clock.c
```diff
@@ -339,57 +361,126 @@ static struct clock enet_clk = { CLK_OPS(ENET, enet, NULL) };
...
+static clk_t *_pll_sys_init(clk_t *clk)
 {
     if (clk->parent == NULL) {
         clk_t *parent = clk_get_clock(clk_get_clock_sys(clk), CLK_MASTER);
         clk_register_child(parent, clk);
         clk->priv = (void *)&clk_regs;
     }
+
+    // After a reset, the system PLL is not active and the clock signal comes
+    // from the 24 MHz oscillator. We do reconfigure anything here, enabling
```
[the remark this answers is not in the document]
axel-h: fixed

libethdrivers/src/plat/imx6/imx6.c
axel-h: > My guess is that the delay is probably to give time for the link speed setting to propagate. There's probably a register to poll for completion of the setting, but I wouldn't know what it would be.

Seems the delay is not needed, https://github.com/seL4/util_libs/pull/83/commits/d4cfc2412e7bf91a9056cc869bc09377a2477cf5 removes it completely eventually.

kent-mcleod:
@axel-h Are you happy for this to be merged now?

libethdrivers/src/plat/imx6/imx6.c
kent-mcleod: Yea I saw this. Because U-boot's model for using ethernet only initializes the hardware directly before use, they probably would see issues with scheduling transmit operations directly after setup without this delay.  Because there's more likely a natural delay from the rest of system setup in the way that the driver's typically used on seL4 this is less of an issue. If you never observe any issue then I'm not going to worry about it.

axel-h:
> @axel-h Are you happy for this to be merged now?

Almost, the change from today removes a superfluous line break in a debug message. Please merge this now.

kent-mcleod:
@axel-h when I try and run the sabre Ethdriver component on qemu (using the picoserver camkes app) I see the following:
```
ocotp_get_mac@ocotp_ctrl.c:236 no MAC in OCOTP for id 0
obtain_mac@imx6.c:629 Failed to get MAC from OCOTP or config
init_device@imx6.c:641 Failed to obtain a MAC
ethif_imx_init_module@imx6.c:939 Failed to initialise the Ethernet driver, code -1
find_compatible_driver_module@io.c:360 Node pointed to by path /soc/aips-bus@2100000/ethernet@2188000 failed to have a driver initialise properly, ignoring
server_init@ethdriver.c:377 Unable to find an ethernet device
```
Do you know why the MAC isn't being read correctly?

axel-h:
QEMU does not emulate the Fuses (OCOTP at 0x021BC000). The driver read zeros there and the driver consider this an unprogrammed fuses, so it expect the MAC to be passed from a config. Unfortunately, QEMU does not support setting OCOTP DATA via `-device loader,addr=0x021BC620,data=0x11223344,data-len=4` so far, so there is no way of emulating OCOTP besides patching it. 
There is no longer a default MAC in the driver lib, instead it expects that the actual driver process to implement the function `get_nic_configuration()` that returns a config structure. See https://github.com/seL4/util_libs/tree/master/libethdrivers/plat_include/imx6/ethdrivers/plat/eth_plat.h

PR #84, opened 2021-04-14
trivial: run style tool

oliver-wm:
Signed-off-by: Oliver Scott <Oliver.Scott@data61.csiro.au>


kent-mcleod:
This PR shouldn't be merged as we generally don't restyle code unless it is being changed.  This is to reduce the barrier for making changes to the style guide.  As a large style change like this creates a large amount of merge-conflict work for any code that is derived from these libraries when it gets updated or contributed back we want to avoid it if at all possible.  But we also want the style guide to be able to be updated without incurring a huge cost from restyling all the code.  So the balance we ended up settling on a while ago was that the style guide needed to be automate-able and that files get restyled as they get updated so that we avoid single commits that make large restyle changes to an entire repository.

oliver-wm:
Cool

axel-h:
I think parts of this could be be merged without causing too much conflicts. One example is switching to "pragma once" everywhere.

PR #85, opened 2021-04-20
Introducing initial bsp for imx8mp-evk board

bvirfollet:
Simple extension for imx8mp-evk BSP
Signed-off-by: Bertrand Virfollet <bvirfollet@silicom.fr>


libethdrivers/CMakeLists.txt
```diff
@@ -63,7 +63,7 @@ mark_as_advanced(
 )
 add_config_library(ethdrivers "${configure_string}")
 
-if("${KernelPlatform}" MATCHES "imx8mq-evk")
+if("${KernelPlatform}" MATCHES "imx8m*-evk")
```
axel-h, line 66: Do we really want to do this or should be better stay with an explicit list for more transparency?

bvirfollet: Hi, I proposed this format as the mix was already done between imx6 and imx8mq. But personally I would have preferred a split into arch / board. Please let me know your point of view about with split view.

Btw, the correct syntaxe is "imx8m.-evk". I'll fix it with a extra commit.

axel-h: Can you try using `if KernelPlatformImx8mq-evk OR KernelPlatformImx8mp-evk OR KernelPlatformImx8mm-evk`? Then we can avoid all the regex pitfalls completely

axel-h: @bvirfollet: do you plan to update the PR?

Issue #86, opened 2021-04-21
iMX8MQ-evk eth driver compilation failure


bvirfollet:
Hi all,

Contrary to what is explained in util_libs/libethdrivers, the inclusion of imx6 platform definitions (util_libs/libethdrivers/src/plat/imx6/uboot/imx-regs.h) is creating compilation errors.

The problem occurs while compiling camkes example picotcp_single_component for imx8mq-evk.
Is the imx8mq-evk ethernet port really supported? Is there a missing commit in the main tree to fix this?
[EthDriver_imx8mq.txt](https://github.com/seL4/util_libs/files/6349660/EthDriver_imx8mq.txt)

From my understanding, it seems that imx-regs.h from uboot should be selectively loaded for imx6 or for imx8m*. Definitions seem really different and should certainly be stored in separated folders, I mean util_libs/libethdrivers/src/plat/imx6 and util_libs/libethdrivers/src/plat/imx8m.

Is it a current action on roadmap? What do you think about it?

axel-h:
You need the first commit from https://bitbucket.hensoldt-cyber.systems/projects/SEOS/repos/sandbox/pull-requests/595/overview, currently this is https://github.com/seL4/util_libs/pull/83/commits/4b3cf7a2031050fccb5472141f96ddc720e32bba but may change whne I rebase it as master moves on.

axel-h:
I've created https://github.com/seL4/util_libs/pull/87 to get this merged quickly

axel-h:
I wonder, could you do  me a favor and test if the patched from https://github.com/seL4/util_libs/pull/87 don't break the i.MX8 support there?

PR #87, opened 2021-04-21
libplatsupport/imx6: add missing include

axel-h:
Fix https://github.com/seL4/util_libs/issues/86, commit taken from https://github.com/seL4/util_libs/pull/83

### DIFF
--- a/libethdrivers/src/plat/imx6/uboot/imx-regs.h
+++ b/libethdrivers/src/plat/imx6/uboot/imx-regs.h
@@ -23,6 +23,7 @@
 
 #pragma once
 
+#include <autoconf.h>
 #include <stdint.h>
 
 #define ARCH_MXC


bvirfollet:
Hi Axel, your branch can build basic camkes socket apps fine for imx8mq (can't test no board) and imx8mp(run fine).
Unfortunatly I still have trouble with the libethdrivers.
Build logs attached
[build_log_merge_4b3cf7.log](https://github.com/seL4/util_libs/files/6358872/build_log_merge_4b3cf7.log)

axel-h:
Seem you have to add you platform there. Currently this file has many place that look like this:
```
#if defined(CONFIG_PLAT_IMX6DQ)
...
#elif defined(CONFIG_PLAT_IMX6SX)
...
#else
#error "unknown i.MX6 SOC"
#endif
```
so it seem you'll have to add `#elif defined(CONFIG_PLAT_IMX8xxx)` for the SOCs od the i.MX8 family.

bvirfollet:
> Seems you have to add your platform there. Currently this file has many place that look like this:
> 
> ```
> #if defined(CONFIG_PLAT_IMX6DQ)
> ...
> #elif defined(CONFIG_PLAT_IMX6SX)
> ...
> #else
> #error "unknown i.MX6 SOC"
> #endif
> ```
> 
> so it seems you'll have to add `#elif defined(CONFIG_PLAT_IMX8xxx)` for the SOCs of the i.MX8 family.Or restructure the define if things are shared between i.MX6 and i.MX8

Hi Axel,
From my point of  view, it seems dangerous to keep mixing imx6 and imx8 definition as they have completely diverged. I think I should better update the content of the libethernet with an updated version extracted from u-boot. That way it would allow better adaptation for other flavors of imx8m boards (m, p and q). 
Don't you think?

axel-h:
We could either aim for a generic i.MX family driver or split this in i.MX6 and i.MX8. I don't know who hacked the i.MX8 support in there and what was the long term plan then. Seems a lot of the stuff is similar and with the driver refactoring I am proposing in another PR it might get even easier. Where do you see the differences exactly besides the slightly different memory layout? And wouldn't this still be largely copy/paste then for the code?

bvirfollet:
> We could either aim for a generic i.MX family driver or split this in i.MX6 and i.MX8. I don't know who hacked the i.MX8 support in there and what was the long term plan then. Seems a lot of the stuff is similar and with the driver refactoring I am proposing in another PR it might get even easier. Where do you see the differences exactly besides the slightly different memory layout? And wouldn't this still be largely copy/paste then for the code?

Hi, 
Finally, I've got this iMX8mp Ethernet driver working. In fact, there isn't much difference with iMX8mq. My trouble was really with the build process and the confusion introduced by with "imx-regs.h" in iMX8 context. This header is misleading for imx8familly boards. 
I started from your branch test/pullreq#83 and only had to derive the appropriate defined(CONFIG_PLAT_IMX8MP_EVK) in libethdrivers and remove the inclusion of "imx-regs.h".

The main difference with imx8mm and imx8mq is the presence of a new ethernet HW with qos facility. I'll try to get it working too before providing a full pull-request for imx8mp.
For the question about split iMX6 and iMX8 code and copy paste problem, I agree there is a lot of it and it much less well organized in libethdriver. I propose to this of the problem otherwise. Wouldn't be interesting to get a framework allowing to get all drivers from u-boot (ideally from Linux) into this libethdrivers and in-fine into user application with seL4 formalism? Do you know if this has been done or started by someone else? Do you know if there is a 'policy' I wouldn't know that have had prevented other dev to make this choice?

oliver-wm (approved):

axel-h:
@bvirfollet I think having something in place that allows using U-Boot driver would be really nice. This would make a nice project to prototype this and use the U-Boot repo instead of cherry-picking source code from U-Boot as it is currently done. Not sure if this falls loosely into the huge domain of the device driver framework (https://www.youtube.com/watch?v=u28M7cMzlB4) then also.

PR #88, opened 2021-05-23
Support both Ethernet ports of i.MX6 SoloX platform

axel-h:
* Support both Ethernet ports of i.MX6 SoloX platform
* fix i.MX6 MAC handling
* improve PLL handling and log messages


libethdrivers/src/plat/imx6/imx6.c
```diff
@@ -706,14 +736,25 @@ static int init_device(imx6_eth_driver_t *dev, const nic_config_t *nic_config)
     /* Initialise the phy library */
     miiphy_init();
     /* Initialise the phy */
+#if defined(CONFIG_PLAT_SABRE) || defined(CONFIG_PLAT_WANDQ)
     phy_micrel_init();
+#elif defined(CONFIG_PLAT_NITROGEN6SX)
+    phy_atheros_init();
```
kent-mcleod, line 742: Is this something that could eventually be queried from part of the device tree?

kent-mcleod (approved):

libethdrivers/src/plat/imx6/imx6.c
axel-h: Yes. In the long run, this should be dynamic. It can either be in the device tree or it could be detected dynamically when setting up the ethernet port, since a PHY is reporting the ID. Not sure what is the best thing here, because the PHY is usually no part of the SoC and it may also change over board revisions. 
Aside from this, `phy_xxx_init()` seems something form U-Boot, how PHY drivers are registered.

axel-h:
Any chance to get this merge before the upcoming release?

PR #97, opened 2021-07-08
Add odroidc4 support

xurtis:
This adds support for the ODroidC4. The ODroidC4 is just a beefier ODroidC2 with some bits moved around. The devices we care about here are all essentially the same.

### DIFF
--- a/libplatsupport/plat_include/odroidc2/platsupport/plat/meson_timer.h
+++ b/libplatsupport/plat_include/odroidc2/platsupport/plat/meson_timer.h
@@ -7,11 +7,7 @@
 #pragma once
 
 #include <platsupport/timer.h>
-
-#define TIMER_BASE      0xc1100000
-#define TIMER_MAP_BASE  0xc1109000
-
-#define TIMER_REG_START   0x2650    // TIMER_MUX
+#include <platsupport/plat/odroid_timer.h>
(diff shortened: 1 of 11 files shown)


nomadeel:
Given that the devices for the C2 and the C4 are the same, would it be better to put the drivers in a common location where both platforms can access them?

xurtis:
> Given that the devices for the C2 and the C4 are the same, would it be better to put the drivers in a common location where both platforms can access them?

The driver implementations are the same, yes, only the headers are different. I could use symlinks for the the .c files for now. I don't think our build system has anything particularly nice for dealing with common devices at the moment.

xurtis:
@nomadeel I've updated this to only introduce new files for address & IRQ differences, everything else is referenced to the odroidc2 using symlinks.

nomadeel:
What you could probably do is this: https://github.com/seL4/util_libs/blob/master/libethdrivers/CMakeLists.txt#L66 and https://github.com/seL4/util_libs/blob/master/libplatsupport/CMakeLists.txt#L111. But even then it's no different than the symlink solution here.

nomadeel:
Regardless of whatever solution you take, it looks good to me so far.

nomadeel (approved):

PR #99, opened 2021-08-16
fix: patches from kent

ridale:
paired with patch in global components.


axel-h:
See https://github.com/seL4/global-components/pull/25

libethdrivers/plat_include/imx6/ethdrivers/plat/eth_plat.h
```diff
@@ -14,11 +14,11 @@
 
 struct enet *get_enet_from_driver(struct eth_driver *driver);
 
-enum {
+enum nic_config_flags_t {
```
axel-h, line 17: why is this needed?

libethdrivers/src/plat/imx6/uboot/imx-regs.h
```diff
@@ -33,6 +33,9 @@
 #define ROMCP_ARB_BASE_ADDR             0x00000000
 #define ROMCP_ARB_END_ADDR              0x000FFFFF
 
+#if defined (CONFIG_PLAT_IMX8MQ_EVK)
+    #define CONFIG_PLAT_IMX6DQ
```
axel-h, line 37: This is semantically broken, if you need things to work on i.MX8, then don't abuse `CONFIG_PLAT_IMX6DQ`, but extend the code to also support `CONFIG_PLAT_IMX8MQ`.

ridale: @kent-mcleod  I don't feel I am the person to defend this? It works, I am all in favor of being able to build code that works, without this patch it doesn't.

kent-mcleod: These changes were the minimum set to be able to get the driver running on the imx8mq.  They would need a round of cleanup before they could be merged.  

This driver used to support the imx8mq platform too (https://github.com/seL4/util_libs/blob/master/libethdrivers/CMakeLists.txt#L68).  I believe when the changes for adding solox were made the support for imx8mq was unintentionally removed.  These changes are a placeholder for what changes would be needed to add imx8mq support back in.

libethdrivers/plat_include/imx6/ethdrivers/plat/eth_plat.h
kent-mcleod: This changes nic_config_flags_t from a global with an enum type to a enum type declaration.  Recent compiler versions (gcc-10) no longer place the global definitions in a COM section and this leads to linker errors. Alternatively this could be changed a different way to add a typedef at the start.

axel-h: I think the typedef is what we need here, this was never supposed to be a global variable.

libethdrivers/src/plat/imx6/uboot/imx-regs.h
axel-h: Can you add a comment in the code that this is a quick hack that is to be replaced then? If there is a i.MX8 board for testing available now I think the time should be take to add this cleanly before it gets merged. Having a `CONFIG_PLAT_IMX8MQ` set up in CMake seems useful in the ling run.

libethdrivers/plat_include/imx6/ethdrivers/plat/eth_plat.h
axel-h: issue looks similar to https://github.com/seL4/camkes-vm/pull/13

PR #101, opened 2021-08-30
Add support for tqma8xqp1gb

bennoleslie:
This add basic platform support for the tqma8xqp1gb.

See: https://github.com/seL4/seL4/pull/520 for more details.

### DIFF
--- a/libplatsupport/plat_include/tqma8xqp1gb/platsupport/plat/clock.h
+++ b/libplatsupport/plat_include/tqma8xqp1gb/platsupport/plat/clock.h
@@ -1,0 +1,15 @@
+/*
+ * Copyright 2021, Data61, CSIRO (ABN 41 687 119 230)
+ * Copyright 2021, Breakaway Consulting Pty. Ltd.
+ *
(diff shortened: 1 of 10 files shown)


kent-mcleod (approved):
There's already a GPT timer based ltimer located in `src/mach/imx/{gpt.c, ltimer.c}` but it uses 2 GPT timers to have one that's always counting up and the other to program timeouts.  Up to you whether you want to use that driver instead of adding this new one, but if both work for sel4test then I'm not to fussed.

bennoleslie:
> There's already a GPT timer based ltimer located in `src/mach/imx/{gpt.c, ltimer.c}` but it uses 2 GPT timers to have one that's always counting up and the other to program timeouts. Up to you whether you want to use that driver instead of adding this new one, but if both work for sel4test then I'm not to fussed.

This is, unfortunately, code I've been sitting on for a while. From memory there was an issue with GPT timer availability. I'd need to deep dive again to work it through. I'd prefer to leave as-is unless there was a strong reason to investigate at this point in time.

PR #102, opened 2021-09-06
libplatsupport,hpet: Fix hpet_start clang issue

kent-mcleod:
On clang-11 setting the ENABLE_CNF flag to enable the hpet was getting
compiled to a byte sized `or` operation and not actually enabling the
timer when running on qemu.

### DIFF
--- a/libplatsupport/src/plat/pc99/hpet.c
+++ b/libplatsupport/src/plat/pc99/hpet.c
@@ -123,7 +123,10 @@ int hpet_start(const hpet_t *hpet)
 
     hpet_timer_t *timer = hpet_get_hpet_timer(hpet->base_addr, 0);
     /* enable the global timer */
-    *hpet_get_general_config(hpet->base_addr) |= BIT(ENABLE_CNF);
+    /* volatile is used here to try and prevent the compiler from satisfying this
+       bitwise operation via byte only reads and writes. */
+    volatile uint64_t *general_config = hpet_get_general_config(hpet->base_addr);
+    *general_config |= BIT(ENABLE_CNF);
 
     /* make sure the comparator is 0 before we turn time0 on*/
     timer->comparator = 0llu;
--- a/libplatsupport/src/plat/pc99/hpet.c
+++ b/libplatsupport/src/plat/pc99/hpet.c
@@ -100,22 +100,22 @@ enum {
 
 static inline uint64_t *hpet_get_general_config(void *vaddr)
 {
-    return (uint64_t*)((uintptr_t)vaddr + GENERAL_CONFIG_REG);
+    return (uint64_t *)((uintptr_t)vaddr + GENERAL_CONFIG_REG);
 }
 
 static inline uint64_t *hpet_get_main_counter(void *vaddr)
 {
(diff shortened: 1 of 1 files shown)


lsf37 (approved):
Content is good. Can you fix the gitlint complaint and add the sign-off?

axel-h:
I'd say this is no a clang issues, but a bug in the code. It's just that clang's aggressive optimization makes this show up. What is the register we are trying to access here. If we do not use the proper accesses, the machine can throw this away. So the comment should be clear here not to imply this to be a clang problem really.

libplatsupport/src/plat/pc99/hpet.c
```diff
@@ -100,30 +100,33 @@ enum {
 
 static inline uint64_t *hpet_get_general_config(void *vaddr)
```
axel-h, line 101: I'd say this return type needs the `volatile` qualifier actually.

kent-mcleod:
> I'd say this is no a clang issues, but a bug in the code. It's just that clang's aggressive optimization makes this show up. What is the register we are trying to access here. If we do not use the proper accesses, the machine can throw this away. So the comment should be clear here not to imply this to be a clang problem really.

I believe that the issue is with qemu considering memory reads and writes < 4 bytes sized to be invalid: https://github.com/qemu/qemu/blob/master/hw/timer/hpet.c#L665

I don't think that this is the same as on actual hardware as it still works there and I'm not aware that x86 restricts byte sized reads and writes to MMIO addresses.

So the comment is only describing what it's trying to do.  It doesn't say that it's a clang bug.

libplatsupport/src/plat/pc99/hpet.c
kent-mcleod: This driver already uses memory fences to ensure correct ordering of memory operations and so adding volatile to all of the memory accesses shouldn't be necessary.

axel-h: but isn't this supposed to be returning a pointer to a 64 device register and not a plain memory location?

kent-mcleod: Yea. When this part of the code was last refactored (https://github.com/seL4/util_libs/commit/478bb400e69df488285907ed07bdfcc44699a3df) all of the volatiles were removed with the assumption that the non-volatile access + memory fences was better because it still allowed the compiler to optimize the code in ways that doesn't introduce bugs.  Unfortunately, now it appears that the compiler is optimizing 32-bit reads and writes into 8-bit in some cases which is potentially against the HPET spec (it says it only supports 32 + 64bit memory operations).  In this case, the memory accesses would need to be forced to stay at 32bit which I believe would require adding inline `asm` accessor functions.  Because that would require rewriting a large part of the driver and a larger amount of work, I opted to using volatile in the one place where the compiler seemed to be performing the bad optimization in a non-critical part of the startup.  I'm not doing any work on this architecture at the moment and I was only helping @lsf37 track down a bug as part of his goal to update the docker containers.

lsf37: I'm happy with it as it as for now. @axel-h if you think this should be refactored in a larger way, we could open an issue for that.

PR #106, opened 2021-09-30
libplatsupport/bcm283x: Unite mailbox implementations

lulu98:
Signed-off-by: Lukas Graber <lukas.graber@hensoldt-cyber.de>


lulu98:
Requires PR https://github.com/seL4/seL4/pull/608 to be merged.

lulu98:
In this PR, we now also remove the dependence on KernelArmMach. New platforms can simply extend the CMakeLists.txt file with an elseif-statement in order to declare a new mach identifier. That way we no longer have to define KernelArmMach in config.cmake in the seL4 kernel repository.

libplatsupport/CMakeLists.txt
```diff
@@ -34,10 +34,20 @@ config_choice(
...
+    set(LibPlatSupportMach "bcm283x")
+else()
+    set(LibPlatSupportMach "")
+endif()
+
 file(
     GLOB
         deps
-        src/mach/${KernelArmMach}/*.c
+        src/mach/${LibPlatSupportMach}/*.c
```
axel-h, line 50: If `LibPlatSupportMach` is empty this becomes `src/mach//*.c`. Seems this problem has also existed before already and this path is just ignored then. Should  be fixed one day. Can we have multiple file(GLOB ...) statement? Then we can simply do this
```
file(
    GLOB
        deps
        src/plat/${KernelPlatform}/*.c
        src/*.c
        src/plat/${KernelPlatform}/acpi/*.c
)

if(NOT ${LibPlatSupportMach} STREQUAL "")
    file(GLOB src/mach/${LibPlatSupportMach}/*.c)
endif()
````

File: libplatsupport/CMakeLists.txt [hunk 2]
```diff
@@ -34,10 +34,20 @@ config_choice(
     "ega;LibPlatSupportX86ConsoleDeviceEGA;LIB_PLAT_SUPPORT_SERIAL_TEXT_EGA;KernelPlatPC99"
 )
 
+if(NOT ${KernelArmMach} STREQUAL "")
```
axel-h, line 37: Maybe we should change the logic to check platform settings first and fall back on the (to be deprecated) `KernelArmMach` only if nothing is set here. 

```
set(LibPlatSupportMach "")
if(KernelPlatformRpi3)
    set(LibPlatSupportMach "bcm283x")
elseif(KernelPlatformRpi4)
    set(LibPlatSupportMach "bcm283x")
else if(NOT ${KernelArmMach} STREQUAL "")
    # falling back to kernel settings is done to keep legacy compatibility.
    set(LibPlatSupportMach "${KernelArmMach}")
endif()
```

libplatsupport/CMakeLists.txt
lulu98: I was only able to make it work when I was first defining another GLOB variable since the * operator does not seem to work with list(APPEND deps src/mach/${LibPlatSupportMach}/*.c). I implemented something similar to https://stackoverflow.com/questions/7533502/how-can-i-merge-multiple-lists-of-files-together-with-cmake.

libplatsupport/CMakeLists.txt
lulu98: Just implemented proposed version, seems to compile.

axel-h (approved):

libplatsupport/CMakeLists.txt
```diff
@@ -34,15 +34,29 @@ config_choice(
     "ega;LibPlatSupportX86ConsoleDeviceEGA;LIB_PLAT_SUPPORT_SERIAL_TEXT_EGA;KernelPlatPC99"
 )
 
+set(LibPlatSupportMach "")
+if(KernelPlatformRpi3)
```
axel-h, line 38: We could merge these to avoid some more redundancy
```
if(KernelPlatformRpi3 OR KernelPlatformRpi4)
    set(LibPlatSupportMach "bcm283x")
...
```

axel-h:
The change look ok to me now. More feedback from other is appreciated if this is the way to go forward.

libplatsupport/CMakeLists.txt
lulu98: Makes sense, corrected it.

nomadeel (approved):
Looks good.

PR #110, opened 2021-11-16
libplatsupport,imx: Update timer drivers to handle more boundary-conditions/edge cases

kent-mcleod:
- When programming timeouts using the ltimer, if the timeout is too soon, ETIME needs to be returned instead of attempting to program a timeout which may end up being and underflowed timer value otherwise
- When reading the upcounter, handle the situation when the counter ticks over between reading it and reading the overflow hardware bit.

### DIFF
--- a/libplatsupport/src/mach/imx/gpt.c
+++ b/libplatsupport/src/mach/imx/gpt.c
@@ -217,9 +217,30 @@ static void gpt_handle_irq(void *data, ps_irq_acknowledge_fn_t acknowledge_fn, v
 
 uint64_t gpt_get_time(gpt_t *gpt)
 {
+    // Rollover of 32-bit counter can happen while we are reading it,
+    // We need to read in the following order (volatile modifier on gpt_map should ensure this):
+    // - GPT_SR[ROV]
+    // - GPT_CNT
+    // - GPT_SR[ROV]
+    // If GPT_SR[ROV] bit is unchanged, then the GPT_CNT read is valid and can be used.
+    // If GPT_SR[ROV] bit has changed, then the rollover happened sometime between the first
+    // read and the third read.  In this case GPT_CNT would have had a value of 0 at some point
+    // during our reads and so use 0.
+    uint32_t rollover_st = gpt->gpt_map->gptsr & BIT(ROV);
     uint32_t low_bits = gpt->gpt_map->gptcnt;
+    uint32_t rollover_st2 = gpt->gpt_map->gptsr & BIT(ROV);
+    if (rollover_st != rollover_st2) {
+        low_bits = 0;
+    }
+
+    // gpt->high_bits is the number of times the timer has overflowed and is managed by this driver.
+    // If GPT_SR[ROV] is set then gpt_handle_irq hasn't had a chance to run yet and gpt->high_bits
+    // is still recording the old number of overflows. We increment our own local copy and leave
+    // the driver copy to be updated by gpt_handle_irq.
+    // This driver requires that the IRQ handler function won't be called while other driver functions
+    // are executing.
     uint32_t high_bits = gpt->high_bits;
-    if (gpt->gpt_map->gptsr) {
+    if (rollover_st2) {
         /* irq has come in */
         high_bits++;
     }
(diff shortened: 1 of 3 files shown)


libplatsupport/src/mach/imx/gpt.c
```diff
@@ -217,9 +217,22 @@ static void gpt_handle_irq(void *data, ps_irq_acknowledge_fn_t acknowledge_fn, v
...
+    uint32_t rollover_st = gpt->gpt_map->gptsr & BIT(ROV);
     uint32_t low_bits = gpt->gpt_map->gptcnt;
+    uint32_t rollover_st2 = gpt->gpt_map->gptsr & BIT(ROV);
+    if (rollover_st != rollover_st2) {
+        low_bits = 0;
+    }
+
     uint32_t high_bits = gpt->high_bits;
-    if (gpt->gpt_map->gptsr) {
+    if (rollover_st2) {
```
axel-h, line 235: Why do we do this check and then increment the value, but why? Can't we just take `high_bits` as it is now? If there had been an overflow, then `low_bits` is 0 here, which seem perfectly fine.

libplatsupport/src/mach/imx/gpt.c
```diff
@@ -217,9 +217,22 @@ static void gpt_handle_irq(void *data, ps_irq_acknowledge_fn_t acknowledge_fn, v
...
+    // Rollover of 32-bit counter can happen while we are reading it,
+    // We need to read in the following order (volatile modifier on gpt_map should ensure this):
+    // rollover_status bit
+    // gptcnt
+    // rollover_status bit.
+    // If rollover_status bit has changed, then the rollover happened sometime between the first
+    // read and the third read.  In this case conclude that the low_bits are 0.
+    uint32_t rollover_st = gpt->gpt_map->gptsr & BIT(ROV);
     uint32_t low_bits = gpt->gpt_map->gptcnt;
+    uint32_t rollover_st2 = gpt->gpt_map->gptsr & BIT(ROV);
```
axel-h, line 229: I wonder if this is enough to catch all corner cases, don't we need something like this:
```
hi_prev = gpt->high_bits
rollover_st = gpt->gpt_map->gptsr & BIT(ROV)
low_bits = gpt->gpt_map->gptcnt
gpt->gpt_map->gptsr & BIT(ROV);
high_bits = gpt->high_bits
if (hi_prev != high_bits) { 
    // there was a high_bits update in between
    low_bits  = 0; 
else if (rollover_st != rollover_st2) && (1 == rollover_st2) { 
    // low bits roll over detected before ISR could update gpt->high_bits
    high_bits++; 
}
```

Then

libplatsupport/src/mach/imx/gpt.c
kent-mcleod: For this particular timer, the counter register is 32bits and the `gpt->high_bits` is a memory region managed by the driver that counts the number of overflows.  When the 32bit counter overflows the ROV bit in the status register is set and an interrupt is generated.  The timer interrupt handler then increments high_bits and clears the status bit.  What's happening here is that the current function effectively snoops the ROV status bit to detect if a rollover has happened but the high_bits hasn't been incremented yet. If so then it increments a local copy of high_bits but doesn't reset the ROV status bit leaving it to the IRQ routine that would run immediately after.

libplatsupport/src/mach/imx/gpt.c
kent-mcleod: The driver's api functions aren't thread safe and the driver requires that what ever is calling into it isn't calling functions concurrently.  Because of this, the value of gpt->high_bits isn't allowed to change within the scope of `gpt_get_time` as it's regular memory and only incremented by the driver.

axel-h: I think that should be in the comment also to explain why this is really sufficient here.

axel-h (approved):

lsf37 (approved):

libplatsupport/src/mach/imx/gpt.c
kent-mcleod: I updated the comments.

lsf37:
The test failure is my fault, haven't set up permissions to secrets yet for this repo. One sec.

lsf37:
> The test failure is my fault, haven't set up permissions to secrets yet for this repo. One sec.

Ok, now it should work.

lsf37:
Ok, the failure is one of those `tx2` dhcp problems. We could re-run everything, but I'd be comfortable merging as is (it'll run again anyway in the main regression).

PR #112, opened 2021-11-17
libethdrivers,imx: Reintroduce ethif_imx6_init

kent-mcleod:
This init function is used by ethif_new_lwip_driver and was previously
removed in an earlier refactor. This commit reintroduces it.

Signed-off-by: Kent McLeod <kent@kry10.com>


libethdrivers/src/plat/imx6/imx6.c
```diff
@@ -788,10 +803,75 @@ static int init_device(imx6_eth_driver_t *dev, const nic_config_t *nic_config)
...
+    }
+
+
+    int ret = init_device(eth_driver, nic_config, plat_config->buffer_addr);
+    if (ret) {
+        ZF_LOGE("Failed to initialise the Ethernet driver, code %d", ret);
+        ret = -ENODEV;
+        return -1;
+    }
+
```
kent-mcleod, line 864: There's a missing return 0 here.

axel-h:
As a follow-up, should we have a CI test that a system using LWIP still builds? Since i.MX6 i supported in QEMU, we could later even add a check that this actually works

kent-mcleod:
> As a follow-up, should we have a CI test that a system using LWIP still builds? Since i.MX6 i supported in QEMU, we could later even add a check that this actually works

This would be the closest: https://github.com/seL4/camkes/pull/13

libethdrivers/plat_include/imx6/ethdrivers/plat/eth_plat.h
```diff
@@ -10,14 +10,14 @@
 #include <ethdrivers/imx6.h>
 #include <utils/attribute.h>
 #include <ethdrivers/raw.h>
-#include "../src/plat/imx6/enet.h"
```
axel-h, line 13: If I remember correctly we had to include this here to address a nasty include problem and make these prototypes available to other C file in our driver:
```
int enet_mdio_read(struct enet *enet, uint16_t phy, uint16_t reg);
int enet_mdio_write(struct enet *enet, uint16_t phy, uint16_t reg, uint16_t data);
```
Could you leave this in for now. I'd make this a separate issue than that should be cleaned up.

axel-h:
I'm fine with merging this when the style issue and the missing return are addressed. Leaving the include file for now is a nice-to-have thing.

libethdrivers/plat_include/imx6/ethdrivers/plat/eth_plat.h
kent-mcleod: It does seem possible to declare just those symbols at the bottom of this file:
```
int enet_mdio_read(struct enet* enet, uint16_t phy, uint16_t reg);
int enet_mdio_write(struct enet* enet, uint16_t phy, uint16_t reg, uint16_t data);
```

Adding a relative include into a private header file breaks if these public headers are installed somewhere else which breaks how I'm using the library.

axel-h: Yes, sounds feasible, I don't remember why we did not do this. Can you put removing this header here in a separate commit, so this is separate from adding the `NIC_CONFIG_NO_CLOCK_SYS` flag or adding the interface again?

axel-h:
@kent-mcleod: do you want to merge this?

PR #114, opened 2021-11-25
Address some outstanding compiler warnings.

kent-mcleod:


libplatsupport/CMakeLists.txt
```diff
@@ -113,6 +113,15 @@ if(KernelPlatformImx8mq-evk OR KernelPlatformImx8mm-evk)
     list(APPEND deps src/plat/imx8m/clock.c)
 endif()
 
+if(KernelPlatPC99)
+    set_source_files_properties(
+        src/plat/pc99/keyboard_vkey.c
+        PROPERTIES
+        COMPILE_FLAGS
+        -Wno-initializer-overrides
```
axel-h, line 121: Can we handle this inside the file with a `#pragma` to narrow down the places where this really matters?

kent-mcleod: I assume it'd require detecting which compiler and emitting the compiler specific pragmas. I'm not that concerned about the small chance this warning hides something new in the future as this file is essentially dead code at this point. My opinion is that scoping the warning to just this file instead of the entire library is already careful enough.

axel-h (approved):

PR #117, opened 2021-12-27
Add Quartz64 support

housel:
This adds support for the Pine64 Quartz64 and other devices based on
the Rockchip RK3566. The platform support is adapted from the
Rockpro64 code, except that the RK356x has A55 cores, and adjusting
for the fact that the ARM Generic Timer is the only on-chip timer
available.

Test with: seL4/seL4#727, seL4_tools#134, seL4/sel4test#83

### DIFF
--- a/libplatsupport/CMakeLists.txt
+++ b/libplatsupport/CMakeLists.txt
@@ -82,7 +82,7 @@ elseif(${KernelArch} STREQUAL "riscv")
     list(APPEND irqchip_modules "-Wl,--undefined=riscv_plic_ptr")
 endif()
 
-if(KernelPlatformQEMUArmVirt)
+if(KernelPlatformQEMUArmVirt OR KernelPlatformQuartz64)
     if(KernelArmExportPCNTUser AND KernelArmExportPTMRUser)
         list(APPEND deps src/arch/arm/generic_ltimer.c)
     endif()
--- a/libplatsupport/src/plat/quartz64/serial.c
+++ b/libplatsupport/src/plat/quartz64/serial.c
@@ -1,0 +1,73 @@
+/*
+ * Copyright 2019, Data61, CSIRO (ABN 41 687 119 230)
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include <string.h>
+#include <stdlib.h>
+#include <platsupport/serial.h>
+#include "../../chardev.h"
+
+#define RHR         0x00
+#define THR         0x00
+#define IER         0x04
+#define LSR         0x14
+#define RHR_MASK    MASK(8)
+#define IER_RHRIT   BIT(0)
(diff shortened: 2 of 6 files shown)


libplatsupport/src/plat/quartz64/serial.c
```diff
@@ -0,0 +1,74 @@
+/*
```
axel-h, line 1: I wonder, is this basically a copy of `src/plat/rockpro64/serial.c`? This could move to `src/mach` then to avoid code duplication.

housel: It is, except the UART addresses are different.

libplatsupport/src/plat/quartz64/serial.c
```diff
@@ -0,0 +1,74 @@
...
+}
+
+int uart_putchar(ps_chardevice_t* d, int c)
+{
+    while (!(*REG_PTR(d->vaddr, LSR) & LSR_TXFIFOE)) {
+        continue;
+    }
+    *REG_PTR(d->vaddr, THR) = c;
+    if (c == '\n' && (d->flags & SERIAL_AUTO_CR)) {
+        uart_putchar(d, '\r');
```
axel-h, line 40: This will print `\n\r` but don't we want to print `\r\n` (CR LF) actually? This bug exists on other platform also and keep spreading with the copy/past of code. This should ne

```
if (c == '\n' && (d->flags & SERIAL_AUTO_CR)) {
        uart_putchar(d, '\r');
}
while (!(*REG_PTR(d->vaddr, LSR) & LSR_TXFIFOE)) {
    /* busy waiting */
}
*REG_PTR(d->vaddr, THR) = c;
```

housel: Added this change

axel-h (approved):
This look ok from my side. I can't test this because I don't have a board. We could add hardware build support in CI to cover this partially at least.

libplatsupport/CMakeLists.txt
```diff
@@ -82,7 +82,7 @@ elseif(${KernelArch} STREQUAL "riscv")
     list(APPEND irqchip_modules "-Wl,--undefined=riscv_plic_ptr")
 endif()
 
-if(KernelPlatformQEMUArmVirt)
+if(KernelPlatformQEMUArmVirt OR KernelPlatformQuartz64)
```
axel-h, line 85: I wonder if we should make this generic for any ARM platform then? And add a ltimer config flag that builds can set to control the behavior explicitly.

axel-h:
@housel Could you address the style issues? The RISC-V error is unrelated, but this PR needs rebasing to fix it.

lsf37:
(rebased and fixed style)

lsf37 (approved):

PR #124, opened 2022-02-18
fix MU_IIR and MU_IER offset addr and enable read interrupt and get char have a -1 returnUpdate serial.c

SaarloosJohn:
fix MU_IIR and MU_IER offset addr and enable read interrupt and get char have a -1 return

Signed-off-by: SaarloosJohn <saarloos@msn.cn>

### DIFF
--- a/libplatsupport/plat_include/bcm2837/platsupport/plat/serial.h
+++ b/libplatsupport/plat_include/bcm2837/platsupport/plat/serial.h
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
- #pragma once
+#pragma once
 
 #define BUS_ADDR_OFFSET             0x7E000000
 #define PADDDR_OFFSET               0x3F000000
--- a/libplatsupport/plat_include/bcm2837/platsupport/plat/serial.h
+++ b/libplatsupport/plat_include/bcm2837/platsupport/plat/serial.h
@@ -12,10 +12,16 @@
 #define UART_BUSADDR                0x7E215000
 
 #define UART_PADDR_0                (UART_BUSADDR-BUS_ADDR_OFFSET+PADDDR_OFFSET)
-/* Broadcom 2835 Peripheral Manual, section 7.5,
- * table "ARM Peripherals interrupts table"
+/* The mini-UART from the Auxiliary peripheral triggers interrupt 29 ("Aux int")
+ * on the VideoCore, which then signals pending GPU interrupts to the ARM core.
+ * Further details can be found in the Broadcom BCM2835/BCM2837 Manual, section
+ * 7.5, table "ARM peripherals interrupts table". The seL4 kernel's interrupt
+ * controller driver creates a linear mapping of all interrupts in the system
+ * (see include/drivers/bcm2836-armctrl-ic.h) with 32 ARM PPIs, 32 basic
+ * interrupts and 64 GPU interrupts. Thus the actual interrupt to be used in
+ * userland is 93 (=32+32+29).
  */
-#define UART_IRQ_0                  (57)
+#define UART_IRQ_0                  (93)
 
 enum chardev_id {
     BCM2837_UART0,
--- a/libplatsupport/src/plat/bcm2837/serial.c
+++ b/libplatsupport/src/plat/bcm2837/serial.c
@@ -38,6 +38,8 @@
 #define MU_LCR_BREAK     BIT(6)
 #define MU_LCR_DATASIZE  BIT(0)
 
+#define MU_IER_ENABLE_READ_INTERRUPT BIT(0)
+
 static void uart_handle_irq(ps_chardevice_t *d UNUSED)
 {
 }
@@ -76,6 +78,8 @@ int uart_init(const struct dev_defn *defn,
     dev->irqs       = defn->irqs;
     dev->ioops      = *ops;
     dev->flags      = SERIAL_AUTO_CR;
+
+    *REG_PTR(dev->vaddr, MU_IER) |= MU_IER_ENABLE_READ_INTERRUPT;
 
     return 0;
 }
--- a/libplatsupport/src/plat/bcm2837/serial.c
+++ b/libplatsupport/src/plat/bcm2837/serial.c
@@ -17,8 +17,8 @@
 #define MU_IO       0x40
 /* When DLAB=1, MU_IIR is a baud rate register.
  * Otherwise IRQ enable */
-#define MU_IIR      0x44
-#define MU_IER      0x48
+#define MU_IER      0x44
+#define MU_IIR      0x48
 #define MU_LCR      0x4C
 #define MU_MCR      0x50
 #define MU_LSR      0x54
--- a/libplatsupport/src/plat/bcm2837/serial.c
+++ b/libplatsupport/src/plat/bcm2837/serial.c
@@ -38,31 +38,30 @@
 #define MU_LCR_BREAK     BIT(6)
 #define MU_LCR_DATASIZE  BIT(0)
 
-static void
-uart_handle_irq(ps_chardevice_t* d UNUSED)
+static void uart_handle_irq(ps_chardevice_t *d UNUSED)
 {
 }
 
-int uart_putchar(ps_chardevice_t* d, int c)
+int uart_putchar(ps_chardevice_t *d, int c)
 {
-    while ( !(*REG_PTR(d->vaddr, MU_LSR) & MU_LSR_TXIDLE) );
+    while (!(*REG_PTR(d->vaddr, MU_LSR) & MU_LSR_TXIDLE));
     *REG_PTR(d->vaddr, MU_IO) = (c & 0xff);
 
     return 0;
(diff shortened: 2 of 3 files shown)


libplatsupport/src/plat/bcm2837/serial.c
```diff
@@ -17,8 +17,8 @@
 #define MU_IO       0x40
 /* When DLAB=1, MU_IIR is a baud rate register.
  * Otherwise IRQ enable */
-#define MU_IIR      0x44
-#define MU_IER      0x48
+#define MU_IIR      0x48
+#define MU_IER      0x44
```
axel-h, line 21: Please keep the numbers in ascending order.

SaarloosJohn: done

axel-h: thank, please squash the commits also

libplatsupport/src/plat/bcm2837/serial.c
```diff
@@ -78,5 +85,7 @@ int uart_init(const struct dev_defn* defn,
     dev->ioops      = *ops;
     dev->flags      = SERIAL_AUTO_CR;
 
+    *REG_PTR(dev->vaddr, MU_IER) |= MU_IER_ENABLE_READ_INTERRUPT;
```
axel-h, line 88: With this change, we are always enabling the interrupt. Are there any side effect with this that we have to consider for existing applications? Wouldn't it be better that is we expose a function here to enable or disable the interrupt, so application can control this? We could track the state via a dedicated bit in `dev->flags` if this is necessary.

SaarloosJohn: I reference other implementations in src/mach, those are all enable interrupt by default, I think may no other effect for this.

libplatsupport/src/plat/bcm2837/serial.c
SaarloosJohn: done

libplatsupport/src/plat/bcm2837/serial.c
```diff
@@ -53,6 +57,9 @@ int uart_putchar(ps_chardevice_t* d, int c)
 
 int uart_getchar(ps_chardevice_t* d UNUSED)
 {
+    if ((*REG_PTR(d->vaddr, MU_IIR) & MU_IIR_INTID) == MU_IIR_HOLD_EMPTY) {
```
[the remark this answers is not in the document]
SaarloosJohn: done

libplatsupport/src/plat/bcm2837/serial.c
axel-h: Good point. Seems there is no specific pattern here, so we can also enable the interrupt by default. Might be more useful in the long run to have it than not to have it

libplatsupport/src/plat/bcm2837/serial.c
```diff
@@ -38,6 +38,11 @@
 #define MU_LCR_BREAK     BIT(6)
 #define MU_LCR_DATASIZE  BIT(0)
 
+#define MU_IER_ENABLE_READ_INTERRUPT BIT(0)
+
+#define MU_IIR_INTID 0x6
```
axel-h, line 43:
```suggestion
#define MU_IIR_INTID_MASK 0x6
```

libplatsupport/src/plat/bcm2837/serial.c
```diff
@@ -38,6 +38,11 @@
...
+#define MU_IIR_HOLD_EMPTY 0x2
```
axel-h, line 44:
```suggestion
#define MU_IIR_INTID HOLD_EMPTY BIT(1)
```

libplatsupport/plat_include/bcm2837/platsupport/plat/serial.h
```diff
@@ -13,9 +13,11 @@
 
 #define UART_PADDR_0                (UART_BUSADDR-BUS_ADDR_OFFSET+PADDDR_OFFSET)
 /* Broadcom 2835 Peripheral Manual, section 7.5,
- * table "ARM Peripherals interrupts table"
+ * table "ARM Peripherals interrupts table", 
+ * 0x7E215000 is Auxiliaries, interrupt is 29,
+ * at kernal/include/drivers/bcm2836-armctrl-ic.h aux irq set to 93.
```
axel-h, line 18: Seems the interrupt 57 was wrong all the time, because it was for the wrong UART and did not take into account the interrupt mapping. Let's rephrase this block to explain how things work and where the numbers actually come from in the RasPi. 

```suggestion
/* The mini-UART from the Auxiliary peripheral triggers interrupt 29 ("Aux int")
 * on the VideoCore, which then signals pending GPU interrupts to the ARM core. 
 * Further details can be found in the Broadcom BCM2835/BCM2837 Manual, section
 * 7.5, table "ARM peripherals interrupts table"). The seL4 kernel's interrupt 
 * controller driver creates a linear mapping of all interrupts in the system
 * (see include/drivers/bcm2836-armctrl-ic.h32) with 32 ARM PPIs, 32 basic 
 * interrupts and 64 GPU interrupts. Thus the actual interrupt to be used in 
 * userland is 93 (=32+32+29).
 */
```

axel-h:
From the code changes this looks good to me now. Can you squash/clean-up the commits in preparation to merge this?

SaarloosJohn:
> From the code changes this looks good to me now. Can you squash/clean-up the commits in preparation to merge this?

axel-h:
Please align the commits with the guidelines, I'd recommend to split this up into these separate commits. Then fixes are cleary separates from changes in the behavior

- libplatsupport,bcm2837: fix UART register offsets
- libplatsupport,bcm2837: fix UART interrupt number
- libplatsupport,bcm2837: enable UART read interrupt
- libplatsupport,bcm2837: return -1 if RX empty

libplatsupport/src/plat/bcm2837/serial.c
```diff
@@ -53,6 +58,9 @@ int uart_putchar(ps_chardevice_t* d, int c)
 
 int uart_getchar(ps_chardevice_t* d UNUSED)
 {
+    if ((*REG_PTR(d->vaddr, MU_IIR) & MU_IIR_INTID_MASK) == MU_IIR_INTID_HOLD_EMPTY) {
+        return -1;
+    }
     while ( !(*REG_PTR(d->vaddr, MU_LSR) & MU_LSR_DATAREADY) );
```
axel-h, line 64: Now that we check `MU_IIR_INTID_HOLD_EMPTY, is the loop around `MU_LSR_DATAREADY` still necessary? Or is there an implicit guarantee that `MU_LSR_DATAREADY` is always set in this case, so we could remove this (or make it a comment).

axel-h: The manual is not really clear here. We could play safe and drop the loop, but still do exactly one check:
```
if (!(*REG_PTR(d->vaddr, MU_LSR) & MU_LSR_DATAREADY)) {
    /* this is not supposed to happen if MU_IIR_INTID says there is data. */
    return -1;
}

SaarloosJohn:
> 

done, Am I do it right?

SaarloosJohn:
> Please align the commits with the guidelines, I'd recommend to split this up into these separate commits. Then fixes are cleary separates from changes in the behavior
> 
> * libplatsupport,bcm2837: fix UART register offsets
> * libplatsupport,bcm2837: fix UART interrupt number
> * libplatsupport,bcm2837: enable UART read interrupt
> * libplatsupport,bcm2837: return -1 if RX empty

done ， is that right?

axel-h:
Seem the style check is still unhappy. The way it works is, that It looks at the modified files only - and it seems this is the first time somebody touched these files since the style rules are in place. The usual way to fix this is to have a first commit that addresses the style complaints. Once the style checker is happy, just add the 4 new commits on top.
The 2nd commit has trailing white space in the commen, this should also be corrected.

SaarloosJohn:
> Seem the style check is still unhappy. The way it works is, that It looks at the modified files only - and it seems this is the first time somebody touched these files since the style rules are in place. The usual way to fix this is to have a first commit that addresses the style complaints. Once the style checker is happy, just add the 4 new commits on top. The 2nd commit has trailing white space in the commen, this should also be corrected.

I make a new commit to fix all the 3 files style.

axel-h:
Thanks. Can you move the commit with the style fixes to be first and make the commit "fix UART interrupt number" not introduce the style violations in the comment it is adding?

SaarloosJohn:
> Thanks. Can you move the commit with the style fixes to be first and make the commit "fix UART interrupt number" not introduce the style violations in the comment it is adding?

ok I will try.

SaarloosJohn:
> Thanks. Can you move the commit with the style fixes to be first and make the commit "fix UART interrupt number" not introduce the style violations in the comment it is adding?

all done

axel-h (approved):
Looks good to me. But I have not tested this on a RasPi myself yet.

PR #126, opened 2022-03-11
Imx8mm compatibility

lucypa:
This is an extension to the imx6 driver to support imx8mq and imx8mm dev boards. 
Ideally the systems will actually have their own files (Todo), and in my opinion there is a tidy up to be done on this folder (I can get onto this in the near future). Perhaps this update should go somewhere else for the time being?


lsf37:
Depending on time frame, we could leave this open as a draft PR until you think it is fully. Or if you think it'll be a bit longer, an in-repo branch here would also be Ok.

lucypa:
@lsf37 It might be a month or two before I'll have time to sort through it again. I also would appreciate feedback from the systems guys on the imx8mm specific changes.

libethdrivers/CMakeLists.txt
```diff
@@ -63,7 +63,7 @@ mark_as_advanced(
 )
 add_config_library(ethdrivers "${configure_string}")
 
-if("${KernelPlatform}" MATCHES "imx8mq-evk")
+if(("${KernelPlatform}" MATCHES "imx8mq-evk") OR ("${KernelPlatform}" MATCHES "imx8mm-evk"))
```
axel-h, line 66: I just wonder, was there any reason we say `MATCHES` here and not `STREQ`? We could simply use the flags here also if this selects well define platforms: `if(KernelPlatformImx8mq-evk OR KernelPlatformImx8mm-evk)`

libethdrivers/src/plat/imx6/enet.c
```diff
@@ -16,8 +16,8 @@
 #include <stdlib.h>
 #include <assert.h>
 
-#ifdef CONFIG_PLAT_IMX8MQ_EVK
-#define CCM_PADDR 0x30380000
+#if (defined(CONFIG_PLAT_IMX8MQ_EVK) || defined(CONFIG_PLAT_IMX8MM_EVK))
+#define CCM_PADDR 0x30380000 // from: root/arch/arm/include/asm/arch-imx8m/imx-regs-imx8mm.h
```
axel-h, line 20: What is "root"?

libethdrivers/src/plat/imx6/imx6.c
```diff
@@ -409,10 +412,11 @@ static void complete_tx(imx6_eth_driver_t *dev)
         /* If this buffer was not sent, we can't release any buffer. */
         if (d->stat & TXD_READY) {
             assert(dev->enet);
+            /* give it another chance */
             if (!enet_tx_enabled(dev->enet)) {
                 enet_tx_enable(dev->enet);
             }
-            return;
+            if (d->stat & TXD_READY) return;
```
axel-h, line 419: please use
```suggestion
            if (d->stat & TXD_READY) {
                return;
            }
```

libethdrivers/src/plat/imx6/imx6.c
```diff
@@ -461,18 +464,21 @@ static void handle_irq(struct eth_driver *driver, int irq)
...
-    if (e & NETIRQ_RXF) {
-        complete_rx(dev);
-        fill_rx_bufs(dev);
-    }
-    if (e & NETIRQ_EBERR) {
-        ZF_LOGE("Error: System bus/uDMA");
-        // ethif_print_state(netif_get_eth_driver(netif));
-        assert(0);
-        while (1);
+    while (e & IRQ_MASK) {
```
axel-h, line 467: Should we merge this as a potential bug fix in a separate PR? Looping to consume all pending interrupt seems a good general pattern and I can't remember any reason not to do it here.

libethdrivers/src/plat/imx6/imx6.c
```diff
@@ -636,6 +643,18 @@ static int init_device(imx6_eth_driver_t *dev, const nic_config_t *nic_config)
...
+    /* Don't attempt to obtain the MAC address for iMX8MQ.
+     *
+     * The code to obtain the MAC address assumes an iMX6 compatible
+     * OCOTP which the iMX8 does not have.
+     *
+     * Instead, we assume that the bootloader has already configured the
+     * hardware MAC address. */
+    uint64_t mac = 0;
+    ZF_LOGI("using MAC configured by bootloader");
+#else
```
axel-h, line 656: I wonder if we should make this a conditional else, so we handle all known platform explicitly and don't have a default behavior? Ir would fail compilation for unknown platforms then and requires explicit action.

libethdrivers/src/plat/imx6/ocotp_ctrl.c
```diff
@@ -211,10 +210,16 @@ uint64_t ocotp_get_mac(struct ocotp *ocotp, unsigned int id)
     uint64_t mac = 0;
 
     switch (id) {
+#if defined(CONFIG_PLAT_IMX8MM_EVK) || defined(CONFIG_IMX8MQ_EVK)
+    case 0:
+        mac = ((uint64_t)((uint16_t)regs->res37[4]) << 32) | regs->res37[0];
```
axel-h, line 215: Please give the registers a name in the definition, reading a MAC from reserved registers is a bit odd.

PR #130, opened 2022-06-05
serial: fix serial port config

abrandnewusername:


libplatsupport/plat_include/pc99/platsupport/plat/serial.h
```diff
@@ -6,6 +6,9 @@
 
 #pragma once
 
+#include <autoconf.h>
```
axel-h, line 9: Is this related to the fix about with the broken numbers? If this just adds missing includes (as a nice-to-have, that kicks in in certain cases?), consider making this separate commit.

kent-mcleod (approved):

libplatsupport/plat_include/pc99/platsupport/plat/serial.h
axel-h: @abrandnewusername could you comment on this?

axel-h:
Seems the style check is unhappy with something in that file. It is not directly related to you change, so we could ignore this for now. But it would be nice if you could add a a commit that does the stale fix.

Concerning the commit comment, fixing the names is not really "trivial", it's fixing a (config) bug after all. So please remove this word.

lsf37:
> Seems the style check is unhappy with something in that file. It is not directly related to you change, so we could ignore this for now. But it would be nice if you could add a a commit that does the stale fix.

Yes, please do add a such a commit. (The policy is to fix style problems in the files we touch, but only the ones we touch, not all files in a repo).

abrandnewusername:
> > Seems the style check is unhappy with something in that file. It is not directly related to you change, so we could ignore this for now. But it would be nice if you could add a a commit that does the stale fix.
> 
> Yes, please do add a such a commit. (The policy is to fix style problems in the files we touch, but only the ones we touch, not all files in a repo).

Thanks for pointing that out, resloved: https://github.com/seL4/util_libs/pull/133

Issue #131, opened 2022-07-18
libpci doesn't handle multiple root complexes.


wom-bat:
On systems where there are PCI or PCIe buses not accessible via a bridge from bus 0, libpci does not find them.

It should allow for multiple root complexes, and set up routing for PCI buses attached other than to the first.

PR #132, opened 2022-07-19
libpci: use "brute force" scan for devices

abrandnewusername:
To handle cases where PCI or PCIe buses are not accessible via
a bridge from bus 0, We use "brute force" scan instead of recursive
scan here. Details of these methods can be found on
https://wiki.osdev.org/PCI#.22Brute_Force.22_Scan.

Signed-off-by: Jingyao Zhou <jingyao.zhou@unsw.edu.au>

This is not the ideal solution, but a workaround for this issue: https://github.com/seL4/util_libs/issues/131

### DIFF
--- a/libpci/src/pci.c
+++ b/libpci/src/pci.c
@@ -129,16 +129,6 @@ static int libpci_add_fun(uint8_t bus, uint8_t dev, uint8_t fun)
 
 static void lib_pci_scan_bus(int bus);
 
-static void lib_pci_scan_fun(int bus, int dev, int fun)
-{
-    libpci_add_fun(bus, dev, fun);
-    if (libpci_read_reg16(bus, dev, fun, PCI_CLASS_DEVICE) == 0x0604) {
-        int new_bus = libpci_read_reg8(bus, dev, fun, PCI_SECONDARY_BUS);
-        ZF_LOGD("%s found additional bus %d from %d %d %d\n", __FUNCTION__, new_bus, bus, dev, fun);
-        lib_pci_scan_bus(new_bus);
-    }
-}
-
 static void lib_pci_scan_dev(int bus, int dev)
 {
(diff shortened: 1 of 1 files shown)


libpci/src/pci.c
```diff
@@ -132,11 +132,6 @@ static void lib_pci_scan_bus(int bus);
 static void lib_pci_scan_fun(int bus, int dev, int fun)
 {
```
wom-bat, line 133: Why not just get rid of this function, and call `libpci_add_fun()` directly from the scan?

wom-bat (changes requested):

kent-mcleod:
> This is not the ideal solution, but a workaround for this issue: #131

Would it be possible to implement this alternate scanning approach as a separate function to maintain the existing functionality for clients that already depend on it?

wom-bat:
>  Would it be possible to implement this alternate scanning approach as a separate function to maintain the existing functionality for clients that already depend on it?
The existing functionality is just an optimisation to avoid scanning everything.   The new code brute-force scans every possible bus, and will pick up any that have been configured by the BIOS.

The existing code is wrong when:
 --- some buses are not attached via bridges but have a separate root complex
 --- the BIOS doesn't enable the bus (the new code won't work for this case either)
 --- where there is more than one PCI domain.

kent-mcleod:
> The existing functionality is just an optimisation to avoid scanning everything. The new code brute-force scans every possible bus, and will pick up any that have been configured by the BIOS.

Is the brute force approach noticeably slower to execute?

wom-bat:
> Is the brute force approach noticeably slower to execute?

Not that I can tell with a single PCI domain (and we don't deal with more than a single PCI domain at the moment).
Of course, to do it properly, you'd parse the DSDT and use the full 4k config space  for each device.

wom-bat (approved):
Looks good to me now.

PR #135, opened 2022-08-03
VM: Support console multiplexing

andybui01:
See: https://github.com/seL4/camkes-vm/pull/34

~~Note: libethdrivers is definitely _not_ the right place to put the headers. I'll separate the virtio headers into their own lib at some point.~~


libethdrivers/include/ethdrivers/virtio/virtio_net.h
```diff
@@ -58,9 +58,11 @@
 #define VIRTIO_NET_S_LINK_UP	1	/* Link is up */
 #define VIRTIO_NET_S_ANNOUNCE	2	/* Announcement is needed */
 
+#define VIRTIO_NET_CONFIG_MAC_SZ 6
+
 struct virtio_net_config {
 	/* The config defining mac address (if VIRTIO_NET_F_MAC) */
-	uint8_t mac[6];
+	uint8_t mac[VIRTIO_NET_CONFIG_MAC_SZ];
```
abrandnewusername, line 65: Might be better to put this in a seperate commit.

kent-mcleod:
Hi @andybui01, if this PR is for merging then are you able to resolve the following items, otherwise the PR should be converted to a draft:
- Resolve the license issues indicated by the License check
- Update the commit messages to be more descriptive (https://docs.sel4.systems/processes/git-conventions.html#commit-messages).  In particular, the commit messages assume too much external context - about the ArmVM project, and so won't make sense for other projects that use these libraries separately from the camkes-vms. 

It would also be helpful for reviewers if you opened an issue to document the improvement or bug that this PR intends to resolve.

PR #136, opened 2022-09-13
virtio: console and vsock support

andybui01:
Add new header files for virtio console and vsock work.


andybui01:
Fails the license check, but the Red Hat licenses are there, untouched.

lsf37:
> Fails the license check, but the Red Hat licenses are there, untouched.

It's not enough for the file to contain any license header, it must contain an `SPDX-Licence-Identifier` declaration and one or  more copyright lines that are machine readable. For the case here, I'd first add a comment

```c
/*
 *  SPDX-License-Identifier: BSD-3-Clause
 */
```

above or below the existing header. If the tool still complains, you might have to include (duplicate) the copyright lines there as well. It probably doesn't like the `(C)` bit and that the year comes at the end, i.e. it'd be

```c
/*
 *  SPDX-License-Identifier: BSD-3-Clause
 * Copyright 2011 Red Hat, Inc.
 * Copyright 2011 Amit Shah <amit.shah@redhat.com> 
 */
```

andybui01:
> It's not enough for the file to contain any license header, it must contain an `SPDX-Licence-Identifier` declaration and one or more copyright lines that are machine readable. For the case here, I'd first add a comment

Noted, thanks!

libethdrivers/CMakeLists.txt
```diff
@@ -117,4 +117,4 @@ string(
     "--undefined=odroidc2_ethernet_ptr"
 )
 
-target_link_libraries(ethdrivers "${DriverModules}")
+target_link_libraries(ethdrivers virtio "${DriverModules}")
```
kent-mcleod, line 120: It's probably better to use:
`target_include_directories(ethdrivers PRIVATE $<TARGET_PROPERTY:virtio,INTERFACE_INCLUDE_DIRECTORIES>)`

This more explicitly only uses the library for the header files it exports.

abrandnewusername (approved):
Looks great to me, please fix the style before merging.

kent-mcleod (approved):

PR #139, opened 2022-10-20
Prelim zynqmp ethdrivers support

chrisguikema:
This PR adds preliminary libethdrivers support for the zynqmp platform.

We have additional work bringing it up to the driver framework, integrating it with the Ethdriver component, and an example webserver application. I just wanted to put this out for preliminary review with the understanding we'll improve on it once the additional features are approved for release.


axel-h:
Could you fix the license identifier to use SPDX, please?

libethdrivers/src/debug.h
```diff
@@ -80,7 +80,7 @@ static inline void print_val(uint8_t *p, int len)
 
 static inline void print_packet(const char *col, void *packet, int length)
 {
-    cprintf(col, "packet 0x%x (%d bytes)", (uint32_t)packet, length);
+    cprintf(col, "packet 0x%x (%d bytes)", (uintptr_t)packet, length);
```
axel-h, line 83:
```suggestion
    cprintf(col, "packet 0x%"PRIxPTR" (%d bytes)", (uintptr_t)packet, length);
```
would be technically correct for `uintptr_t`. May not matter practically, though.

chrisguikema:
Alright, I ended up just reworking these commits to pull the updated zynq7000 libethdrivers, and then applied all the style patches so the checks would pass. 

This should be good to review and merge, as it doesn't prevent the builds. And then DornerWorks will make some additional PRs to utilize the library when the code is approved for release.

libethdrivers/src/plat/zynqmp/io.h
```diff
@@ -0,0 +1,155 @@
...
+#define __arch_getl(addr)         *((volatile uint32_t*)(addr))
+#define __arch_getw(addr)         *((volatile uint16_t*)(addr))
+#define __arch_getb(addr)         *((volatile uint8_t*)(addr))
+
+#define __arch_putl(val, addr)    *((volatile uint32_t*)(addr)) = val
+#define __arch_putw(val, addr)    *((volatile uint16_t*)(addr)) = val
+#define __arch_putb(val, addr)    *((volatile uint8_t*)(addr)) = val
+
+
+//#define __raw_writel(...) writel(__VA_ARGS__)
```
axel-h, line 18: consider removing or add a comment why this is commented out.

libethdrivers/src/plat/zynqmp/io.h
```diff
@@ -0,0 +1,155 @@
+/*
```
axel-h, line 1: This seem a generic util lib that is not ZynqMP specific. Maybe move it up a bit, so it can also be used by other code?

chrisguikema: This was copied directly from the zynq7000 libethdrivers.

A future PR could reformat to a `mach` directory for the common code between the `zynqmp` and `zynq7000` ethdrivers. But I think a more incremental approach would be better here.

libethdrivers/src/plat/zynqmp/io.h
chrisguikema: Copied directly from zynq7000 libethdrivers. The only design decisions are in the second commit, where the code was modified to work with the zynqmp.

kent-mcleod:
> Alright, I ended up just reworking these commits to pull the updated zynq7000 libethdrivers, and then applied all the style patches so the checks would pass.

So U-Boot has a single driver for zynq7000 and zynqmp, zynq_gem.c? Is the new driver in this PR synced with upstream U-Boot, or is it just based off the forked one in zynq7000/ from this library?

kent-mcleod (approved):

chrisguikema:
> Is the new driver in this PR synced with upstream U-Boot, or is it just based off the forked one in zynq7000/ from this library?

`zynq_gem.c` was copied from the forked one in `libethdrivers/src/plat/zynq7000`

PR #143, opened 2022-11-09
libplatsupport: add UART for qemu-riscv-virt platform

axel-h:
Test with: seL4/seL4#818, seL4/sel4test#82


lsf37 (commented):
@axel-h can you please fix up the gitlint + style failures? The content looks fine from my side.

@Ivan-Velickovic if you're happy with the content as well, we can merge the entire set.

lsf37 (approved):

axel-h:
> Have you tested the UART driver? My understanding is that seL4test does not make use of it, I just want to double check that get char and put char work.

I've copied this from another driver. But I will run an explicit test then.

axel-h:
I've tested the driver now with a customized QEMU that has multiple UARTs.

PR #148, opened 2023-02-14
Fix compilation warning about implicit conversion of a negative value

ar-cetitec:
It is converted to an unsigned type: an "int -1" to "unsigned long".

### DIFF
--- a/libcpio/src/cpio.c
+++ b/libcpio/src/cpio.c
@@ -208,7 +208,7 @@ const void *cpio_get_file(const void *archive, unsigned long len, const char *na
         if (error) {
             return NULL;
         }
-        if (cpio_strncmp(header_info.filename, name, -1) == 0) {
+        if (cpio_strncmp(header_info.filename, name, (unsigned long)(-1)) == 0) {
             break;
         }
         len = cpio_len_next(len, header, header_info.next);


libcpio/src/cpio.c
```diff
@@ -208,7 +208,7 @@ const void *cpio_get_file(const void *archive, unsigned long len, const char *na
         if (error) {
             return NULL;
         }
-        if (cpio_strncmp(header_info.filename, name, -1) == 0) {
+        if (cpio_strncmp(header_info.filename, name, ~0ul) == 0) {
```
axel-h, line 211: I wonder if an explicit cast might be a bit mote intuitive:

```suggestion
        if (cpio_strncmp(header_info.filename, name, (unsigned long)(-1)) == 0) {
```
Otherwise, if we have `ULONG_MAX` here:
```suggestion
        if (cpio_strncmp(header_info.filename, name, ULONG_MAX) == 0) {
```

ar-cetitec: I was worried about some toolchain missing the "limits.h", but the repository already uses it, so `ULONG_MAX` it is.

ar-cetitec: (according to [godbolt](godbolt.org) there is something called "ppci 0.5.5 compiler on CPython", which does not have it)

ar-cetitec:
Updated commit description to better match common commit style in the repository.

libcpio/src/cpio.c
kent-mcleod: Yea, we can't use limits.h because this library is used in a bootloader that doesn't provide a limits.h.

kent-mcleod: You can see this in the failing compilation test: https://github.com/seL4/util_libs/actions/runs/4183101374/jobs/7247936534

I think explicitly casting to an unsigned long is more appropriate in this case.

ar-cetitec: Updated. Though personally, I consider `~0ul` as concise as possible.

axel-h (approved):
Looks good to me. Just the git commit message linter seems still unhappy that the title is too long, just use "libcpio: fix conversion of -1 to ulong"

ar-cetitec:
> Looks good to me. Just the git commit message linter seems still unhappy that the title is too long, just use "libcpio: fix conversion of -1 to ulong"

Done

axel-h:
Could you rebase this to master also?

ar-cetitec:
> Could you rebase this to master also?

oops, too late. Thanks

PR #152, opened 2023-04-21
libutils: compile_time_assert to support various standards

elmankku:
`static_assert` and `_Static_assert` were introduced to C11. However gcc and clang support `_Static_assert` for earlier versions. C23 deprecates `_Static_assert`, `static_assert` is changed to a keyword and will be removed from `assert.h`. 

Change compile_time_assert to take these into account.

See also: https://github.com/seL4/seL4_projects_libs/pull/96

### DIFF
--- a/libutils/include/utils/compile_time.h
+++ b/libutils/include/utils/compile_time.h
@@ -8,4 +8,10 @@
 
 #include <assert.h>
 
+/* 201112L indicates C11 */
+#if __STDC_VERSION__ < 201112L && !defined(__cplusplus)
+/* Use gcc extension for older C standards. */
+#define compile_time_assert(name, expr) _Static_assert((expr), #name)
+#else
 #define compile_time_assert(name, expr) static_assert((expr), #name)
+#endif


libutils/include/utils/compile_time.h
```diff
@@ -8,4 +8,10 @@
 
 #include <assert.h>
 
+#if !defined(static_assert) && \
```
axel-h, line 11: Let's add a comment here `/* 201112L indicated c11 */`. Might be extra verbose, but it's a nice reference to demystify the numer.

elmankku: Sure, will add.

libutils/include/utils/compile_time.h
```diff
@@ -8,4 +8,11 @@
 
 #include <assert.h>
 
+/* 201112L indicates C11 */
+#if !defined(static_assert) && \
```
axel-h, line 12: I also wonder, why do we check `!defined(static_assert)` actually? We should drop this, if we have c11 then use `_Static_assert` as defined by the standard. The check should be 

```
#if defined(__cplusplus)
/* ToDo: what shall we do for C+++? */
#elif __STDC_VERSION__ >= 201112L /* c11 or higher */
#define compile_time_assert(name, expr) _Static_assert((expr), #name)
#else
#define compile_time_assert(name, expr) static_assert((expr), #name)
#endif
```

elmankku: Yes, `!defined(static_assert)` is useless. Forgot it there when I tried slightly different. I'll remove that.

I would keep
```
#if __STDC_VERSION__ < 201112L && !defined(__cplusplus)
define compile_time_assert(name, expr) _Static_assert((expr), #name)
#else
#define compile_time_assert(name, expr) static_assert((expr), #name)
#endif
```

because it supports:
- C11 and newer with `static_assert()`, including C23 where `static_assert()` is becoming a keyword and `_Static_assert()` deprecated
- C99/gnu99 and other older C standards on gcc/clang with `_Static_assert`
- C++11 and newer with `static_assert()`

The check requires `!defined(___cplusplus)` because `_Static_assert()` gcc extension is not supported for C++ and couldn't find a way to do this without imposing limitations for the macros (see below).

I wouldn't recommend including the fallback
```
#else
#define compile_time_assert(name, expr) typedef int __assert_failed_##name[(expr) ? 1 : -1] __attribute__((unused));
#endif
```
because that won't compile if the `name` argument is a string. For example:
```
compile_time_assert("This would produce error with fallback", 0); // doesn't work with the fallback
compile_time_assert(this_would_work_with_the_fallback, 0);
```

There are a few places where strings are being used.

@axel-h what do you think?

kent-mcleod (approved):

axel-h (approved):
Looks good to me

PR #154, opened 2023-05-02
libplatsupport: avoid clang warning

axel-h:
Avoid warning on clang (v11.0.1): "`'cold' attribute only applies to functions`". 

The warning make be think no optimization is happening at all with clang based on this attribute. Anything happening would either be a lucky coincidence oder some other heuristic kicks in. 

This fix allows us to build with clang, as we handle all warnings as errors by default.

### DIFF
--- a/libutils/include/utils/assume.h
+++ b/libutils/include/utils/assume.h
@@ -26,6 +26,15 @@
  * used. A perfect example is error handling code. This gives the compiler a
  * light hint to deprioritise optimisation of this path.
  */
+
+#ifdef __clang__
+
+/* Avoid warning: 'cold' attribute only applies to functions */
+#define COLD_PATH() do { } while(0)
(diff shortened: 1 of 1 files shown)


kent-mcleod:
> This fix allows us to build with clang, as we handle all warnings as errors by default.

Where is this creating a build failure?

axel-h:
> Where is this creating a build failure?

It is used here: https://github.com/seL4/camkes-tool/blob/master/libsel4camkes/include/camkes/error.h#L230

Seem that is the only place, `HOT_PATH()` seem the be used nowhere at all. So maybe removing this macro is the better option anyway?

kent-mcleod:
> Seem that is the only place, `HOT_PATH()` seem the be used nowhere at all. So maybe removing this macro is the better option anyway?

Yea I think that would be fine.

PR #157, opened 2023-05-08
libplatsupport: separate drivers from platforms

axel-h:
Many platform use the same peripherals, so code duplication can be avoided if we provide generic drivers.


kent-mcleod:
> Many platform use the same peripherals, so code duplication can be avoided if we provide generic drivers.

The original thought was to use a similar strategy as the kernel does for querying a platform device tree and a configuration file for what compatibility strings to match on and then pulling in source files based on these match strings.

axel-h:
So, I'd say this PR is a step into that direction then?

PR #158, opened 2023-05-08
trivial: Fix clang warning about register width

kent-mcleod:
Changing the input argument to be word-sized avoids the inline assembly overwriting more of the register on 64-bit architectures and appeases the compiler.

Fixes warnings like:
```
/tmp/tmp.3CplW6S5Lu/projects/util_libs/libplatsupport/src/arch/arm/delay.c:60:29: warning: value size does not match register size specified by the constraint and modifier [-Wasm-operand-widths]
        "    bpl 1b" : "+r"(cycles));
                            ^
/tmp/tmp.3CplW6S5Lu/projects/util_libs/libplatsupport/src/arch/arm/delay.c:55:23: note: use constraint modifier "w"
        "    subs %0, %0, #1;"  /* 12 */
                      ^~
                      %w0
```

### DIFF
--- a/libplatsupport/src/arch/arm/delay.c
+++ b/libplatsupport/src/arch/arm/delay.c
@@ -32,7 +32,7 @@ static unsigned long cpufreq_hint = DEFAULT_CPUFREQ;
 
 void ps_udelay(unsigned long us);
 
-static void ps_do_cycle_delay(int32_t cycles)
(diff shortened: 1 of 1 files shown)


kent-mcleod:
Duplicate of https://github.com/seL4/util_libs/pull/153

PR #160, opened 2023-05-17
libplatsupport: fix build warnings

axel-h:
Use proper format specifiers to avoid build warnings "`warning: format specifies type 'unsigned long long' but the argument has type 'uint64_t' (aka 'unsigned long') [-Wformat]`" on 64-bit platforms. Seen with clang, which seems a bit more picky than gcc.


axel-h:
Now the style checker should be happy also.

kent-mcleod (approved):

PR #161, opened 2023-05-25
libethdrivers: make zynqmp driver usable

axel-h:
- fix compatibility string. Seems this changed in at some point in the device tree? Or did this ever work, because that seems the zynq7000 string? 
- add zynqmp driver to the list of symbols kept, so the linker will not throw this out.

@chrisguikema I'm a bit confused actually, how do things work for you without these changes?

### DIFF
--- a/libethdrivers/CMakeLists.txt
+++ b/libethdrivers/CMakeLists.txt
@@ -118,6 +118,7 @@ string(
     "-Wl"
     "--undefined=tx2_ether_qos_ptr"
     "--undefined=zynq7000_gem_ptr"
+    "--undefined=zynqmp_gem_ptr"
     "--undefined=imx_fec_ptr"
     "--undefined=odroidc2_ethernet_ptr"
 )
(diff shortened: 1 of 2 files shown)


abrandnewusername (approved):
Looks great to me.

chrisguikema:
@axel-h Long story short, a lot of our customers require us to OPSEC each commit we want to open-source, which takes a long time. The two commits we released were ones we had approval for, and we've been using the zynqmp ethdrivers way before the driver framework was released. The goal was to have another series of commits made available that make the zynqmp usable, with an example platform, but the OPSEC process is taking a lot longer than anticipated

axel-h:
Can we merge this or is something blocking this still?

libethdrivers/CMakeLists.txt
```diff
@@ -118,6 +118,7 @@ string(
     "-Wl"
     "--undefined=tx2_ether_qos_ptr"
     "--undefined=zynq7000_gem_ptr"
+    "--undefined=zynqmp_gem_ptr"
```
Indanz, line 121: I don't understand why this is needed for a library. And if it's not a library, then why not discard it if unused?

Indanz (approved):

libethdrivers/CMakeLists.txt
kent-mcleod: Those symbols get put into a special named section. At runtime the section gets iterated over when looking for compatible drivers. Because the symbols aren't directly linked the object file may not end up being linked in if the symbols aren't explicitly marked undefined.

PR #162, opened 2023-05-30
libfdt: Add 32/64-bit fdt_appendprop_uint()

hlyytine:
Adds a convenience function that reduces the number of 'if' statements in the code, since #cells required is given as a parameter.

### DIFF
--- a/.stylefilter
+++ b/.stylefilter
@@ -1,7 +1,7 @@
 # Copyright 2019, Data61, CSIRO (ABN 41 687 119 230)
 # SPDX-License-Identifier: BSD-2-Clause
 
-libfdt
+libfdt/*
 liblwip/default_opts/lwipopts.h
 # Imported sources from zf_log
 libutils/include/utils/zf_log.h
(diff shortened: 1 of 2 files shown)


hlyytine:
Seems that style check fails, even though `libfdt` shouldn't be checked against seL4 style rules. I just can't remember what file I am supposed to add to bypass the check for this?

hlyytine:
Note that seL4/sel4_projects_libs#108 is also required.

hlyytine:
> It has `libfdt` in there, but his does not work. Using `libfdt/*` seems to be the proper expression to use.

That did the trick, thanks.

axel-h:
> > It has `libfdt` in there, but his does not work. Using `libfdt/*` seems to be the proper expression to use.
> 
> That did the trick, thanks.

Could you make the stylefilter change a separate commit?

axel-h:
> Note that [seL4/seL4_projects_libs#108](https://github.com/seL4/seL4_projects_libs/pull/108) is also required (just found out you had accepted [seL4/camkes-vm#107](https://github.com/seL4/camkes-vm/pull/107), but the other repo does not have reviews)

For this PR here it is not needed, it just depends on it. Or did I ger something wrong?

hlyytine:
> For this PR here it is not needed, it just depends on it. Or did I ger something wrong?

Never mind, it was just too early morning for me.

kent-mcleod (approved):

PR #165, opened 2023-06-26
hifive: Check if overflow IRQ is actually pending

Indanz:
Timer may have been reset, or the overflow may be already handled by pwm_get_time().

This fixes sel4test [issue 91](https://github.com/seL4/sel4test/issues/91).


lsf37 (approved):

kent-mcleod (approved):

kent-mcleod:
I can confirm that the timer test does run for the correct duration now.

PR #167, opened 2023-08-21
Add pine64 star64 platform

canarysnort01:
This adds a timer and serial driver to libplatsupport.
Co-developed with @Ivan-Velickovic.

**Purpose**: Ready to merge.
**Context**: Needed to allow sel4test to run on star64.
**Testing performed**: Tested with WIP sel4test for the star64 platform on pinetab-v hardware.


Ivan-Velickovic:
It should be noted that due to there being no current documentation on the JH7110 timer, details such as the register map are taken from the JH7110 timer Linux kernel patch which can be found here: https://patchwork.kernel.org/project/linux-riscv/patch/20230627055313.252519-3-xingyu.wu@starfivetech.com/.

libplatsupport/src/plat/star64/serial.c
```diff
@@ -0,0 +1,67 @@
...
+    while ((*REG_PTR(d->vaddr, UART_LSR) & BIT(0)));
+    return *REG_PTR(d->vaddr, UART_THR);
+}
+
+int uart_putchar(ps_chardevice_t *d, int c)
+{
+    while ((*REG_PTR(d->vaddr, UART_LSR) & UART_LSR_THRE) == 0);
+
+    /* Add character to the buffer. */
+    *REG_PTR(d->vaddr, UART_THR) = c & 0x7f;
```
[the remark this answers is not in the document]
canarysnort01: I tested printing a UTF-8 emoji and it came through, so I guess it can print 8 bits.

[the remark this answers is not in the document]
canarysnort01: Fixed

libplatsupport/src/plat/star64/serial.c
```diff
@@ -0,0 +1,67 @@
...
+    if (c == '\n' && (d->flags & SERIAL_AUTO_CR)) {
```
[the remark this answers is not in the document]
canarysnort01: Fixed

axel-h:
It looks good, could you squash the commits?

libplatsupport/src/plat/star64/serial.c
```diff
@@ -0,0 +1,68 @@
...
+
+    /* Add character to the buffer. */
+    *REG_PTR(d->vaddr, UART_THR) = c;
+
+    return c;
+}
+
+static void uart_handle_irq(ps_chardevice_t *dev)
+{
+    /* This UART driver is not interrupt driven, there is nothing to do here. */
```
axel-h, line 41: Since the UARTs are used as output only, interrupt support does not matter practically. One day, some generic seL4 console demo (tetris, nethack ...)  would be nice to get read interrupts added also...

canarysnort01: IIUC based on other examples, to support this the only thing I need to do is to enable the interrupt generation on the device on init, and ack the interrupt to the device in this function?

axel-h: basically yes.

canarysnort01: I will add that.

But it's not clear to me how a consumer of this API would effectively deal with interrupts. It looks like they are supposed to register and wait for the interrupt themselves, and then tell the driver when the interrupt occurred, without getting any further information about the interrupt from the driver.

On this device there could be various reasons for the interrupt and I don't see a way for a client to determine why it was generated.

canarysnort01: Added

axel-h: If I remember correctly, uart_handle_irq() is just the abstraction function to ack the interrupt. The lib's user must still attach to the interrupt and call the uart_read() wrapper to drain the FIFO first. An existing app that does this should work out of the box on the Pine64.

libplatsupport/src/plat/star64/timer.c
```diff
@@ -0,0 +1,149 @@
...
+void starfive_timer_handle_irq(starfive_timer_t *timer)
+{
+    assert(timer);
+    assert(timer->regs);
+
+    timer->value_h += 1;
+
+    while (timer->regs->intclr & STARFIVE_TIMER_INTCLR_BUSY) {
+        /*
+         * Hardware will not currently accept writes to this register.
```
axel-h, line 54:
```suggestion
         * Hardware does not accept writes to this register.
```

libplatsupport/plat_include/star64/platsupport/plat/timer.h
```diff
@@ -0,0 +1,71 @@
...
+    uint32_t intmask;
+} starfive_timer_regs_t;
+
+typedef struct {
+    volatile starfive_timer_regs_t *regs;
+    /*
+     * Stores the number of times the continuous counter timer has elapsed and started over.
+     * This allows us to count to a higher number than allowed by the hardware.
+     */
+    uint32_t value_h;
```
axel-h, line 61: Should this be declared volatile also, as it could be access concurrently, and we also want to prevent the compiler from doing some optimization that would move the accesses around.

libplatsupport/plat_include/star64/platsupport/plat/timer.h
```diff
@@ -0,0 +1,71 @@
...
+#define STARFIVE_TIMER_CHANNEL_1_IRQ 70
+#define STARFIVE_TIMER_CHANNEL_2_IRQ 71
+#define STARFIVE_TIMER_CHANNEL_3_IRQ 72
+
+/* This information comes from the DTS file */
+#define STARFIVE_TIMER_BASE 0x13050000
+#define STARFIVE_TIMER_REGISTER_WINDOW_LEN_IN_BYTES 0x10000
+#define STARFIVE_TIMER_TICKS_PER_SECOND 0x16e3600
+
+#define STARFIVE_TIMER_MAX_TICKS 0xffffffff
```
axel-h, line 29: Do we have `UINT32_MAX` here?
```suggestion
#define STARFIVE_TIMER_MAX_TICKS  UINT32_MAX
```

libplatsupport/plat_include/star64/platsupport/plat/timer.h
```diff
@@ -0,0 +1,71 @@
...
+#define STARFIVE_TIMER_CHANNEL_REGISTERS_LEN_IN_BYTES 0x40
+#define STARFIVE_TIMER_CHANNEL_0_IRQ 69
+#define STARFIVE_TIMER_CHANNEL_1_IRQ 70
+#define STARFIVE_TIMER_CHANNEL_2_IRQ 71
+#define STARFIVE_TIMER_CHANNEL_3_IRQ 72
+
+/* This information comes from the DTS file */
+#define STARFIVE_TIMER_BASE 0x13050000
+#define STARFIVE_TIMER_REGISTER_WINDOW_LEN_IN_BYTES 0x10000
+#define STARFIVE_TIMER_TICKS_PER_SECOND 0x16e3600
```
axel-h, line 27: Please add a comment at leat, that this is 24 MHz, so the strange hex number is a bit more clear.

libplatsupport/src/plat/star64/timer.c
```diff
@@ -0,0 +1,149 @@
...
+{
+    assert(timer);
+    assert(timer->regs);
+
+    /* the timer value counts down from the load value */
+    uint64_t value_l = (uint64_t)(STARFIVE_TIMER_MAX_TICKS - timer->regs->value);
+    uint64_t value_h = (uint64_t)timer->value_h;
+
+    /* Include unhandled interrupt in value_h */
+    if (timer->regs->intclr == 1) {
```
axel-h, line 72: How is the register intclr defined, are we really checking the whole register value here, or just the bit 0, because the other bits are reserved? I could not find any document that describes the timer resisters, seem this was never published and this is just taken from some other code?

libplatsupport/src/plat/star64/timer.c
```diff
@@ -0,0 +1,149 @@
...
+        value_h += 1;
```
axel-h, line 73: Should we read timer->regs->value again now, as the interrupt may have arrived after we read it before?

libplatsupport/src/plat/star64/timer.c
```diff
@@ -0,0 +1,149 @@
...
+}
+
+uint64_t starfive_timer_get_time(starfive_timer_t *timer)
+{
+    assert(timer);
+    assert(timer->regs);
+
+    /* the timer value counts down from the load value */
+    uint64_t value_l = (uint64_t)(STARFIVE_TIMER_MAX_TICKS - timer->regs->value);
+    uint64_t value_h = (uint64_t)timer->value_h;
```
axel-h, line 69: Do we have to take into account concurrent around `timer->value_h` here? So better read `high`, `lo`, and `high` again. If high it differs, read `low` again.

libplatsupport/src/plat/star64/timer.c
canarysnort01: I couldn't find any documentation either other than what was with some code. I did test that this was set as shown when there was an unhandled interrupt. I can add the mask to be defensive.

axel-h: Well, without documentation, it's a well-educated guess that it's following the usual way, where just bit 0 matters. Bitght be safer to add a comment, that it's unclear how exactly this register works and give the source where this code is copied from.

canarysnort01: To be clear, I didn't copy the source, just used it as documentation.

libplatsupport/plat_include/star64/platsupport/plat/timer.h
Ivan-Velickovic: I'm confused, these drivers aren't multithreaded right? How could there be concurrent access?

axel-h: Driver ware not multi-threaded, but they have an ISR that could interrupt us any time and update this value. Using `volatile` might not prevent the race completely, but could makes the windows even smaller.

libplatsupport/src/plat/star64/timer.c
Ivan-Velickovic: Isn't that what line 68 is doing? I'm confused

libplatsupport/src/plat/star64/timer.c
Ivan-Velickovic: The time is always going to be changing because you have to consume time to read the register so would reading the registers again actually help?

libplatsupport/plat_include/star64/platsupport/plat/timer.h
Ivan-Velickovic: I don't quite understand why `volatile` will help here, this is internal state that is just like a normal C variable?

libplatsupport/src/plat/star64/serial.c
Ivan-Velickovic: I think this is resolved.

libplatsupport/src/plat/star64/timer.c
Ivan-Velickovic: I've added a comment in `timer.h` that points to the Linux patches.

Ivan-Velickovic:
I've tried to address all the comments now, hopefully we can merge this.

PR #168, opened 2023-09-01
[FIX] Fix bcm pl011 UART

FelixSchladt:
I tried using the existing driver for data transmissions (with a hardware rng) and had the issue that without enabled FIFO, no interrupts are received.

As the note states, enabling the FIFO seems to cause some other trouble with the SerialServer.
I think this should be discussed, what makes the most sense and maybe a solution which fits all can be found.


libplatsupport/src/mach/bcm/pl011_uart.c
```diff
@@ -96,16 +98,18 @@ static inline void pl011_uart_disable_fifo(ps_chardevice_t *dev)
     r->lcrh &= ~LCRH_FEN;
 }
 
-static inline void pl011_uart_enable_rx_irq(ps_chardevice_t *dev)
+static inline void pl011_uart_enable_irqs(ps_chardevice_t *dev)
```
axel-h, line 101: Now that we enable multiple interrupt, can you add a comment here what each enabled interrupt reports. The acronyms are not intuitive.

libplatsupport/src/mach/bcm/pl011_uart.c
```diff
@@ -128,7 +132,7 @@ static int pl011_uart_cr_configure(ps_chardevice_t *dev)
     uint32_t val = r->cr;
 
     val |= CR_TXE; // Transmit enable
-    val |= CR_RXE; // Teceive enable
+    val |= CR_RXE; // Receive enable
```
axel-h, line 135: please make typo fixes a separate commit.

libplatsupport/src/mach/bcm/pl011_uart.c
FelixSchladt: done

libplatsupport/src/mach/bcm/pl011_uart.c
```diff
@@ -227,14 +230,14 @@ static int pl011_uart_configure(ps_chardevice_t *dev)
      *
      */
     // Enable FIFO
-    //pl011_uart_enable_fifo(dev);
+    pl011_uart_enable_fifo(dev);
+
+    // Enable RX interrupt
```
axel-h, line 235:
```suggestion
    // Enable interrupts
```

axel-h:
Just to clarify, it seem to me that when you enable the receive timeout interrupt and the FIFO, then the serial server should work again. You get an interrupt either when the FIFO has a certain level or - due to the receive timeout - when there is data in the FIFO, but no additional data is received for "a 32-bit period". So we should be able to merge this without breaking thins. I wonder if should still make the interrupt behavior configurable via the `flags` field of `ps_chardevice_t`, so we can keep the old behavior by default and enable the new behavior only if explicitly selected.

libplatsupport/src/mach/bcm/pl011_uart.c
FelixSchladt: I updated the commits

Indanz (approved):
All look like reasonable fixes.

As the old code is known not to work and this probably does, I think we should just merge this.

Indanz:
The Gitlint and License Checks are stuck, but I can't re-trigger them. I also don't have an option to bring the branch up-to-date, so I guess a manual merge needs to be done.

FelixSchladt:
I will try to rebase it on monday

Indanz:
So I looked at the Broadcom UART documentation, but it doesn't explain what the receive time-out does. Does it trigger all the time or does it only trigger after receiving a character and nothing else arrives?

FelixSchladt:
> So I looked at the Broadcom UART documentation, but it doesn't explain what the receive time-out does. Does it trigger all the time or does it only trigger after receiving a character and nothing else arrives?

As far as I understand, the receive timeout only triggers after a certain period of no incoming data

Indanz:
> As far as I understand, the receive timeout only triggers after a certain period of no incoming data

Does that mean you get an interrupt all the time? Because that would make it "work", but not in the right way.

libplatsupport/src/mach/bcm/pl011_uart.c
```diff
@@ -227,14 +230,14 @@ static int pl011_uart_configure(ps_chardevice_t *dev)
      *
      */
     // Enable FIFO
-    //pl011_uart_enable_fifo(dev);
+    pl011_uart_enable_fifo(dev);
```
Indanz, line 233: I'm not convinced that the problem mentioned in the comment is properly addressed. I am afraid that enabling FIFOs is just a work-around for an unfixed bug, probably caused by the original's code assumptions of the current state of the UART or initialisation.

Have you tried fully configuring the UART and setting all registers, and clearing any pending IRQs and errors at init?

FelixSchladt: I tested the current state again today and it is working fine.

Sending single bytes is not an issue. 
For testing I did setup a simple uart echo server which is interrupt triggered.

Indanz: Please print something out at every receive interrupt and check that you don't get any when no input is given.

Or keep a counter and print out the counter value too.

PR #169, opened 2023-09-07
Update mach timer.h to use axi instead of amba

chrisguikema:
Test with: https://github.com/seL4/seL4/pull/1102

### DIFF
--- a/libplatsupport/mach_include/zynq/platsupport/mach/timer.h
+++ b/libplatsupport/mach_include/zynq/platsupport/mach/timer.h
@@ -23,10 +23,17 @@
 #define IRQS_PER_TTC 3
 
 #ifdef CONFIG_PLAT_ZYNQMP
+#ifdef CONFIG_ZYNQMP_PETALINUX_AXI
+#define TTC0_PATH "/axi/timer@ff110000"
+#define TTC1_PATH "/axi/timer@ff120000"
+#define TTC2_PATH "/axi/timer@ff130000"
+#define TTC3_PATH "/axi/timer@ff140000"
+#else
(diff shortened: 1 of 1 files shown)


axel-h:
> Updates to the kernel device tree forces this change.

Can you extend this comment comment with some version details, so it gets more transparent where this happened. Guessing from your branch name it seems peta linux 2021.1 needs this?

chrisguikema:
> > Updates to the kernel device tree forces this change.
> 
> Can you extend this comment comment with some version details, so it gets more transparent where this happened. Guessing from your branch name it seems peta linux 2021.1 needs this?

Done.

I updated the kernel PR to add the option to select a specific petalinux version

axel-h:
> I updated the kernel PR to add the option to select a specific petalinux version

This seem to allows some transition then. But I wonder it such a config option is really needed practically. We could just move on and support a newer version (where the commit comment says, what version this is). If the expect everybody to move to this version anyway, there is not much gain keeping support for older version. Or you have to keep your own fork then for legacy support.
I wonder if a long term solution must be factoring this out of the C code into some config file, so we can have profiles.

chrisguikema:
> But I wonder it such a config option is really needed practically. 

I agree with you. But a theme of the PRs I've made is not breaking existing features/support. While updating to 2021.1 now would _probably_ be okay, once zcu102 configurations are released to camkes-vm-apps, then future updates to the kernel would be breaking. This feels like a good compromise, where we can just release a new version without introducing any breaking changes.

chrisguikema:
Not needed based on kernel PR feedback.

PR #170, opened 2023-09-11
Change to support BSD sed

alwin-joshy:
In GNU `sed`, the suffix is optional for the `-i` argument, but is mandatory for the BSD `sed` shipped with MacOS. This change checks whether the host is an Apple system and if so, supplies a suffix.

### DIFF
--- a/libpicotcp/CMakeLists.txt
+++ b/libpicotcp/CMakeLists.txt
@@ -45,14 +45,14 @@ if(LibPicotcp)
         " -I$<JOIN:$<TARGET_PROPERTY:muslc,INTERFACE_INCLUDE_DIRECTORIES>, -I>"
     )
 
-    if (CMAKE_C_COMPILER_ID STREQUAL "Clang")
+    if(CMAKE_C_COMPILER_ID STREQUAL "Clang")
         list(APPEND compile_options "${CMAKE_C_COMPILE_OPTIONS_TARGET}${CMAKE_C_COMPILER_TARGET}")
         set(C_COMPILER clang)
     else()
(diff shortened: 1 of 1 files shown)


alwin-joshy:
The other two changes are unrelated but were required to get style checks passing.

libpicotcp/CMakeLists.txt
```diff
@@ -72,6 +72,11 @@ if(LibPicotcp)
     add_custom_target(get_muslc)
     add_dependencies(get_muslc muslc)
 
+    set(sed_inplace_suffix "")
```
Ivan-Velickovic, line 75: Need a comment explaining why the behaviour is needed on macOS.

lsf37: Yes, a comment here would be nice

libpicotcp/CMakeLists.txt
```diff
@@ -72,6 +72,11 @@ if(LibPicotcp)
...
+    if(CMAKE_HOST_APPLE)
+        set(sed_inplace_suffix "''")
+    endif()
```
lsf37, line 78: How does GNU `sed` treat `-i ''`? Would that happen do work on both BSD and GNU?

alwin-joshy: Unfortunately not, `-i''` does seem to work though.

lsf37: Ok. Might be worth mentioning in the comment, because I can see us trying that again in the future.

lsf37 (approved):

Ivan-Velickovic (approved):

PR #172, opened 2024-01-11
morello: Add support for QEMU platform

heshamelmatary:
This was tested on a baseline AArch64 Morello/QEMU without CHERI.


axel-h:
Do you expect this to match the hardware, so we'd put this in a `morello` folder and not `morello-qemu`

libplatsupport/src/plat/morello-qemu/serial.c
```diff
@@ -0,0 +1,74 @@
...
+    return ch;
+}
+
+int uart_putchar(ps_chardevice_t *d, int c)
+{
+    while ((*REG_PTR(d->vaddr, UARTFR) & PL011_UARTFR_TXFF) != 0);
+
+    *REG_PTR(d->vaddr, UARTDR) = c;
+    if (c == '\n' && (d->flags & SERIAL_AUTO_CR)) {
+        uart_putchar(d, '\r');
```
axel-h, line 41: This writes `\n\r`, but it should be `\r\n`

axel-h:
Since this created another PL011 driver, it could be put on top of https://github.com/seL4/util_libs/pull/157 to reduce the duplication a bit.

heshamelmatary:
> Do you expect this to match the hardware, so we'd put this in a `morello` folder and not `morello-qemu`

Not quire. PL011's address and IRQ are different. Also hardware has extra clock, pl011, and timer (SP804) drivers. Unfortunately, I don't think drivers can be currently shared in util_libs.

heshamelmatary:
> Since this created another PL011 driver, it could be put on top of #157 to reduce the duplication a bit.

I believe that's the right direction, but are there plans to merge #157 soon? If not, we could just merge this and then remove later not to block https://github.com/seL4/seL4/issues/1158 ?

libplatsupport/src/plat/morello-qemu/serial.c
heshamelmatary: I just copied it from `libplatsupport/src/plat/qemu-arm-virt/serial.c`

axel-h: It is broken there also, see 970c79da for a fix

PR #174, opened 2024-01-28
github: retire pr-queue action

lsf37:
This is the last pace this action was still running. It has been largely ignored, so we should retire it.


lsf37:
This should be merged before seL4/ci-actions#321

axel-h (approved):

PR #176, opened 2024-01-31
libutils: add kernel word size related CTZ macro

colorglass:
Add kernel word size related macros `CxZ_WORD()` for ctz and clz builtin func. And also add `CxZLL()` macro for 64bits data size.
And for the macro CTZL(), which is used for `unsigned long` type, I found a error use in  sel4test timer dirver: https://github.com/seL4/sel4test/blob/0a488b92bc7a08acc14d49524d1f22656f286627/apps/sel4test-driver/src/timer.c#L53
```
......
void handle_timer_interrupts(driver_env_t env, seL4_Word badge)
{
    int error = 0;
    while (badge) {
        seL4_Word badge_bit = CTZL(badge);
......
```
The `seL4_Word` type can be 64bits, but CTZL() recives `unsigned long` argument. It can be changed to `CTZ_WORD()` to fix the problom.

### DIFF
--- a/libutils/include/utils/builtin.h
+++ b/libutils/include/utils/builtin.h
@@ -14,6 +14,8 @@
 #define CLZ(x) __builtin_clz(x)
 #define CTZL(x) __builtin_ctzl(x)
 #define CLZL(x) __builtin_clzl(x)
+#define CTZLL(x) __builtin_ctzll(x)
+#define CLZLL(x) __builtin_clzll(x)
 #define FFS(x) __builtin_ffs(x)
 #define FFSL(x) __builtin_ffsl(x)
 #define OFFSETOF(type, member) __builtin_offsetof(type, member)
@@ -25,8 +27,22 @@
 
 #if CONFIG_WORD_SIZE == 32
 #define BSWAP_WORD(x) __builtin_bswap32(x)
+compile_time_assert(long_is_32bits, sizeof(long) == 4)
+#define CTZ_WORD(x) CTZL(x)
+#define CLZ_WORD(x) CLZL(x)
+
 #elif CONFIG_WORD_SIZE == 64
 #define BSWAP_WORD(x) __builtin_bswap64(x)
+compile_time_assert(long_long_is_64bits, sizeof(long long) == 8)
+#define CTZ_WORD(x) CTZLL(x)
+#define CLZ_WORD(x) CLZLL(x)
+
+#else
+    /*
+     *  This library aims to be agnostic of the "word" concept, so don't raise an
+     *  error if CONFIG_WORD_SIZE is unsupported or not set. We just don't provide
+     *  the macros at all in this case.
+     */
 #endif
 
 /* The UNREACHABLE macro is used in some code that is imported to Isabelle via


libutils/include/utils/builtin.h
```diff
@@ -29,6 +31,18 @@
 #define BSWAP_WORD(x) __builtin_bswap64(x)
 #endif
 
+#if CONFIG_WORD_SIZE == 32
```
axel-h, line 34: let's group the variants together and issue a warning if `CONFIG_WORD_SIZE` is invalid.
```
    #if CONFIG_WORD_SIZE == 32
    #define BSWAP_WORD(x) __builtin_bswap32(x)
    #define CTZ_WORD(x) CTZL(x)
    #define CLZ_WORD(x) CLZL(x)
    #elif CONFIG_WORD_SIZE == 64
    #define BSWAP_WORD(x) __builtin_bswap64(x)
    #define CTZ_WORD(x) CTZLL(x)
    #define CLZ_WORD(x) CLZLL(x)
    #else
    #error "unsupported CONFIG_WORD_SIZE"
    #endif
```

libutils/include/utils/builtin.h
```diff
@@ -29,6 +31,18 @@
...
+#define CTZ_WORD(x) CTZL(x)
+#elif CONFIG_WORD_SIZE == 64
+#define CTZ_WORD(x) CTZLL(x)
```
axel-h, line 37: Actually, on 64-bit systems, it is quite common that `long` is 64-bit. So we could also go with `#define CTZ_WORD(x) CTZL()` in both cases. Note that In `libutils/include/utils/arith.h` this lib already uses `CTZL()`.

colorglass: You are right, but I tested it using both x64 and arm64 `msvc` 19.30 and get that `long` is 32bits size. So I suggest that we use the more likely selection for `word` size type.

axel-h: Oh - if `long` is 32-bit on a 64-bit architecture, I'd expect this to break a lot more things...

lsf37: `long` is definitely 64 on the compilers that are supported for the 64 bit architectures that are supported.

colorglass:
I searched C99 standard [ISO/IEC 9899:1999](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf), and didn't find the size define of `long`, only got some information indicates that it should hold the number limits defined in `limits.h`. 
And I find that the size of `long` in gcc was defined by a internal macro `LONG_TYPE_SIZE` [Layout of Source Language Data Types](https://gcc.gnu.org/onlinedocs/gccint/Type-Layout.html), which defines that `long` type in gcc has the default **one word size**.
Finally I write the simple test file to print the result of `sizeof(long)`, compile and run and get the result is 4 within my visual studio 2019.
So I can confirm that the practical size of the `long` is depended on the compiler itself. The most generally used compiler is gcc and one word size `long` becomes a de facto standard.
Please forgive my attention to details, and since there is no builtin_ctz() function in msvc, we can just apply CTZL() to word size variable with no change to this file.

PR #182, opened 2024-05-15
libplatsupport: Add support for imx8mp-evk

nomadeel:
The i.MX 8M Plus Evaluation Kit is similar to the other platforms in the i.MX 8M family and shares the same serial and timer peripherals.


Indanz (approved):

libplatsupport/CMakeLists.txt
```diff
@@ -132,7 +133,13 @@ add_config_library(platsupport "${configure_string}")
 
 add_library(platsupport EXCLUDE_FROM_ALL ${deps})
 
-if(KernelPlatImx8mq OR KernelPlatformImx8mm-evk)
+if(KernelPlatImx8mq OR KernelPlatformImx8mm-evk OR KernelPlatformImx8mp-evk)
+    if(KernelPlatformImx8mp-evk)
+        # For the imx8mp, the DTS is in a different format compared to those of
+        # the imx8mm and imx8mq, there are some constants in the headers that
+        # point to the proper path
+        target_compile_definitions(platsupport PRIVATE USE_IMX8MP_GPT_PATH)
```
Indanz, line 141: You could use `#ifdef CONFIG_PLAT_IMX8MP_EVK` below and avoid this, that would be more in line with the existing code. If it gets messy in the future we can always add defines, also for serial. But if you prefer it this way then that's fine with me too.

PR #189, opened 2024-11-29
Initial i.MX93 support

Indanz:
Works with the ARM generic timer when `KernelArmExportPCNTUser` and `KernelArmExportPTMRUser` are enabled. The Low Power Timer driver is in progress and will be added soon.

### DIFF
--- a/libplatsupport/CMakeLists.txt
+++ b/libplatsupport/CMakeLists.txt
@@ -86,7 +86,7 @@ else()
     message(FATAL_ERROR "Unsupported KernelArch '${KernelArch}'")
 endif()
 
-if(KernelPlatformQEMUArmVirt OR KernelPlatformQuartz64)
+if(KernelPlatformQEMUArmVirt OR KernelPlatformQuartz64 OR KernelPlatformIMX93)
     if(KernelArmExportPCNTUser AND KernelArmExportPTMRUser)
         list(APPEND deps src/arch/arm/generic_ltimer.c)
     endif()
--- a/libplatsupport/CMakeLists.txt
+++ b/libplatsupport/CMakeLists.txt
@@ -35,6 +35,18 @@ config_choice(
     "ega;LibPlatSupportX86ConsoleDeviceEGA;LIB_PLAT_SUPPORT_SERIAL_TEXT_EGA;KernelPlatPC99"
 )
 
+config_choice(
+    LibPlatSupportLPTMRclock
+    LIB_PLAT_SUPPORT_LPTMR_CLOCK
+    "Which input clock to use for the Low Power Timer\
+    32.768 kHz -> High accuracy, but low precision.\
+    24 MHz -> Higher precision, but lower accuracy. Synchronous with the scheduling clock."
+    "32kHz;LibPlatSupportLPTMRclock24MHz;LIB_PLAT_SUPPORT_LPTMR_CLOCK_32kHZ;KernelPlatformIMX93"
+    "24MHz;LibPlatSupportLPTMRclock32kHz;LIB_PLAT_SUPPORT_LPTMR_CLOCK_24MHz;KernelPlatformIMX93"
+)
+
+mark_as_advanced(CLEAR LibPlatSupportX86ConsoleDevice LibPlatSupportLPTMRclock)
+
 set(LibPlatSupportMach "")
 if(KernelPlatformRpi3 OR KernelPlatformRpi4)
     set(LibPlatSupportMach "bcm")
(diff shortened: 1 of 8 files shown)


Xaphiosis:
> Works with the ARM generic timer when `KernelArmExportPCNTUser` and `KernelArmExportPTMRUser` are enabled. The Low Power Timer driver is in progress and will be added soon.

As far as we know, you added support for the low-power timer already. Is it only the case that this comment is out of date now?

lsf37:
> As far as we know, you added support for the low-power timer already. Is it only the case that this comment is out of date now?

The corresponding commit is still missing on this branch.

lsf37 (approved):
This looks good from my side. It would be nice to add the LPTR even just for convenience. Happy for it to go into this PR or a separate one.

Indanz:
More detail in my mail. To summarize, at least on the MCIMX93-SOMB2 module the 32 kHz clock is significantly out-of-sync with the 24 MHz clock (assuming they didn't put a 25 MHz crystal on by mistake), most likely because the 24 MHz clock is inaccurate by about 2.5%. This makes the SCHED0011 test fail because it relies on the clocks to be mostly in sync.

Another issue is that taking a timestamp takes 0.5 us, which is extremely slow.

Basically we have four choices:

1) Don't bother with the LPTMR driver.
2) Add the LPTMR using the 32 kHz clock: sel4test will fail on this board.
3) Add LPTMR using the 24 MHz clock: sel4test will pass, timing will probably be very inaccurate.
4) Add LPTMR using 32 kHz clock, but use Arm Generic Timer for timestamping when `CONFIG_EXPORT_PCNT_USER` is enabled. This should make make the tests pass and be practically useful. But would cost more time to implement.

Currently I have a driver with a define at the top that chooses between the two clock sources.

Indanz:
I pushed the LPTMR, I made the clock source a config option, with the 24 MHz source as default to avoid breaking sel4test.

Indanz:
The clock mystery also has been solved:

I measured the clocks with a borrowed oscilloscope and that proofs that the 24MHz is fine, but the 32 kHz one is garbage. They put a 32.768 kHz crystal on the dev board instead of a 32 kHz one, probably because it's cheaper. That explains the 2.4% difference I see. They even documented it in MCIMX93EVKUM.pdf and the schematics, but I missed it before.

I'm going to complain about this to NXP. Those dev boards are expensive, so you'd expect them to not cut corners like this. If they do things like this, you can as well get a much cheaper, but probably higher quality i.MX93 board from someone else instead.

Indanz:
Never mind, I'm just stupid: The 32kHz crystal is supposed to be 32.768 kHz. Updating the driver to that and change to the 32 kHz cock source as default. I've been misled by 32kHz dividing one second in a nice round nanosecond number, but that's not what the crystal value was chosen for, it was chosen to divide one second in a power-of-two number of ticks.

Indanz:
Could someone review my timer driver code? I'm not very familiar with the util_libs' driver framework, it's a bit confusing. Other timer drivers seem to be either over complicating things, or I'm missing something.

lsf37 (approved):
Happy with the updates and with the figured out mystery of the 32 kHz clock. The driver code as such looks good to me, but I can't say I'm an expert on these.

libplatsupport/CMakeLists.txt
```diff
@@ -86,7 +98,7 @@ else()
     message(FATAL_ERROR "Unsupported KernelArch '${KernelArch}'")
 endif()
 
-if(KernelPlatformQEMUArmVirt OR KernelPlatformQuartz64)
+if(KernelPlatformQEMUArmVirt OR KernelPlatformQuartz64 OR KernelPlatformIMX93)
     if(KernelArmExportPCNTUser AND KernelArmExportPTMRUser)
         list(APPEND deps src/arch/arm/generic_ltimer.c)
```
kent-mcleod, line 103: This makes me a bit uncomfortable because `KernelArmExportPTMRUser` is a very insecure option. The kernel doesn't do anything at the moment to protect or context switch these timer registers meaning that any user level TCB can read/write to them. In the qemu platform I didn't think it was too much cause for concern because it's only a development platform.
But for imx93 we should look to add an ltimer driver that doesn't use the generic timer.
(Or we more loudly ask for kernel support for allowing user level to securely using the physical timer/virtual timer that the kernel doesn't use). There's a few platforms that don't have a usable platform timer (like zynqmp) and would benefit from using the generic timer.
The situation's even worse on riscv because there's no secondary arch core timer. (Here the kernel could technically provide a virtual timer service on MCS quite easily.)

Indanz: I did add a proper ltimer which isn't based on the platform timer, see third commit. Only if the platform timer is exported to user space it will be used, otherwise the LPTMR will be used. I left it in case there were problems with the LPTMR (it's very slow to read).

kent-mcleod: I see. I missed that there was `ltimer_default_init()` at the bottom of that new module and that there were ifdef guards at the top of the file.

PR #191, opened 2025-03-06
cva6: add missing chardev include for dev_defn

midnightveil:
This was previously giving a build warning regarding it only being defined within this file itself.

```c
/util_libs/libplatsupport/src/mach/cva6/serial.c:12:28: warning: ‘struct dev_defn’ declared inside parameter list will not be visible outside of this definition or declaration
   12 | int uart_init(const struct dev_defn *defn,
      |                            ^~~~~~~~
```

### DIFF
--- a/libplatsupport/src/mach/cva6/serial.c
+++ b/libplatsupport/src/mach/cva6/serial.c
@@ -4,10 +4,10 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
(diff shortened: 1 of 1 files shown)


lsf37 (approved):

lsf37:
Still need to wrap the commit message body at 72 to pass the Gitlint check, but otherwise looks fine to me.

PR #193, opened 2025-03-31
libplatsupport: export a config if no ltimer

midnightveil:
Previously this would be a linker error later.
Now this should hopefully make it clear by removing the prototypes and so should create a compile error (if flags are set correctly).

https://github.com/seL4/sel4bench/pull/55

Style failures are unrelated.


kent-mcleod (approved):

lsf37 (approved):
Looks good. Have added a style fixup commit for the file that we have touched.

PR #196, opened 2025-05-29
ltimer: indicate no support for more platforms

midnightveil:
These were taken from the sel4test Sel4testHaveTimer config option.

~~**Do not merge** until https://github.com/seL4/util_libs/pull/195 has been merged (as it will break sel4test for more platforms)~~ ok now

### DIFF
--- a/libplatsupport/CMakeLists.txt
+++ b/libplatsupport/CMakeLists.txt
@@ -48,7 +48,13 @@ config_choice(
 mark_as_advanced(CLEAR LibPlatSupportX86ConsoleDevice LibPlatSupportLPTMRclock)
 
 # Some platforms don't have a platform timer.
-if(KernelPlatformCheshire)
+if(
+    (KernelPlatformQEMUArmVirt AND NOT (KernelArmExportPCNTUser AND KernelArmExportPTMRUser))
+    OR KernelPlatformRocketchip
+    OR KernelPlatformRocketchipZCU102
+    OR KernelPlatformCheshire
+    OR (SIMULATION AND (KernelArchRiscV OR KernelArchARM))
+)
     set(LibPlatSupportHaveTimer OFF)
 else()
     set(LibPlatSupportHaveTimer ON)
--- a/libplatsupport/include/platsupport/ltimer.h
+++ b/libplatsupport/include/platsupport/ltimer.h
@@ -325,7 +325,7 @@ static inline void ltimer_us_delay(ltimer_t *timer, uint64_t microseconds)
  * with calling the ltimer functions inside the callback, the ltimer interface
  * functions are not reentrant.
  */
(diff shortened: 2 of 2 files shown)


Indanz (approved):

libplatsupport/CMakeLists.txt
```diff
@@ -48,7 +48,13 @@ config_choice(
 mark_as_advanced(CLEAR LibPlatSupportX86ConsoleDevice LibPlatSupportLPTMRclock)
 
 # Some platforms don't have a platform timer.
-if(KernelPlatformCheshire)
+if(
+    (KernelPlatformQEMUArmVirt AND NOT (KernelArmExportPCNTUser AND KernelArmExportPTMRUser))
+    OR KernelPlatformRocketchip
+    OR KernelPlatformRocketchipZCU102
+    OR KernelPlatformCheshire
+    OR (SIMULATION AND (KernelArchRiscV OR KernelArchARM))
```
lsf37, line 56: @midnightveil for some reason this line breaks the camkes timer component build. For all Arm platforms we're getting:

```
  /github/workspace/projects/global-components/components/TimeServer/src/time_server.c: In function ‘post_init’:
  /github/workspace/projects/global-components/components/TimeServer/src/time_server.c:230:13: error: call to ‘ltimer_default_init’ declared with attribute error: no ltimer support for this platform
    230 |     error = ltimer_default_init(&ltimer, io_ops, time_server_ltimer_handle, NULL);
        |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

(e.g. https://github.com/seL4/camkes-tool/actions/runs/15923333523/job/45177979627)

If I remove this line, everything builds and works fine. Either SIMULATION does not mean what it says or it is on for those builds even though they are hardware images. It does seem to work if I explicitly provide `-DSIMULATION=OFF` on the command line.

Are we actually sure that it is true for all Arm platforms that qemu does not provide an ltimer device? I thought that some of them have fairly extensive device simulation (but maybe not timer, I have not checked).

lsf37: `SIMULATE` does just seem to be on for those builds. It is actually generating a `simulate` script (which then says the platform is not supported for simulation).

So the right fix seems to be to make sure SIMULATE is not on for those builds, but it makes me wonder if we're braking other builds without intending to.

lsf37: Ok, so it does intentionally switch on SIMULATION by default in [easy-settings.cmake](https://github.com/seL4/camkes/blob/master/easy-settings.cmake)
```
set(SIMULATION ON CACHE BOOL "Try and use simulable features")
```

So I'm not entirely sure how to proceed. Do we actually need to guard SIMULATION here in util_libs, or is it just to fail early?

PR #198, opened 2025-07-02
ltimer: enable timer for QEMU Arm arch timers

midnightveil:
The if statements were not strictly equivalent and broke the camkes timer component builds:

https://github.com/seL4/util_libs/pull/196/files#r2178857087

Ref: https://github.com/seL4/sel4test/pull/142


Indanz (commented):
I think it would be better to remove the `SIMULATION` part of the check than shuffling this bit around like this.

PR #202, opened 2025-11-03
Add Support for Banana Pi BPI-F3

akifejaz:
related to https://github.com/seL4/seL4/pull/1535

This driver is based on the implementations in [PR #167](https://github.com/seL4/util_libs/pull/167) and [PR #190](https://github.com/seL4/util_libs/pull/190), with necessary updates, such as address and configuration changes applied according to the current DTS for this platform added here https://github.com/seL4/seL4/blob/772179e4e639fec1c48a01c69ea6cea144cab02a/tools/dts/bananapi-f3.dts

### DIFF
--- a/libplatsupport/mach_include/spacemit/platsupport/mach/timer.h
+++ b/libplatsupport/mach_include/spacemit/platsupport/mach/timer.h
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2025, 10xEngineers
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+#pragma once
+
+/* The K1 SoC has multiple timers, the 'Timer1' peripheral is used
+ * by this driver. Each timer peripheral consists of three individual
+ * counters/timers, confusingly named Timer 0, Timer 1 and Timer 2.
+ * Below Timer 0 is called TIMER0 and is used as a counter and Timer 1
+ * is called TIMER1 and used for timeouts.
+ * K1 SoC spec : https://developer.spacemit.com/documentation?token=AZsXwVrqaisaDGkUqMWcNuMVnPb
+ */
+
+#include <stdint.h>
+#include <stdbool.h>
+#include <stddef.h>
+
+#define SPACEMIT_TIMER_BASE      0xD4014000
+#define APB_CLK_BASE             0xD4015000
+#define SPACEMIT_TIMER_SIZE      0x2000u
+#define SPACEMIT_TIMER0_IRQ      23u // timer0_1_irq: Timer 0 IRQ of Timer1
+#define SPACEMIT_TIMER1_IRQ      24u // timer0_2_irq: Timer 0 IRQ of Timer2
+#define SLOW_CLOCK false
+
+#if SLOW_CLOCK
+#define SPACEMIT_TIMER_TICKS_PER_SECOND  32768u
+#define K1_TCCR_CS0_VALUE                0x1u   /* slow clock */
+#else
+#define SPACEMIT_TIMER_TICKS_PER_SECOND  12800000u
+#define K1_TCCR_CS0_VALUE                0x0u   /* fast clock */
+#endif
+
+#define SPACEMIT_NUM_TIMERS 3
+#define SPACEMIT_TIMER_MAX_TICKS UINT32_MAX
+#define APBC_TIMERx_CLK_RST_OFFSET  0x34u
+#define SPACEMIT_TIMERS_CNT_OFFSET  0x90u
+
+typedef struct {
+    uint32_t tcer;         // Timer Count Enable Register
+    uint32_t tcmr;         // Timer Count Mode Register
+    uint32_t tcrr;         // Timer Count Restart Register
+    uint32_t tccr;         // Timer Clock Control Register
+
+    uint32_t tmr[3][4];    // Timer Match Register (TMRx)
+    uint32_t tplvr[4];     // Timer Preload Value Register (TPLVRx)
+    uint32_t tplcr[4];     // Timer Preload Control Register (TPLCRx)
+
+    uint32_t tier[4];      // Timer Interrupt Enable Register (TIERx)
+    uint32_t ticr[4];      // Timer Interrupt Clear Register (TICLRx)
+    uint32_t tsr[4];       // Timer Status Register (TSRx)
+
+    uint32_t tcnt[4];      // Timer Count Register (TCRx)
+} spacemit_timer_regs_t;
+static_assert(offsetof(spacemit_timer_regs_t, tcnt) == SPACEMIT_TIMERS_CNT_OFFSET,
+              "struct spacemit_timer_regs_t has incorrect layout");
+
+typedef struct {
+    volatile spacemit_timer_regs_t *regs;
+    uint8_t  timer_n;
+    uint32_t value_h;
+    uint32_t *vaddr;
+} spacemit_timer_t;
+
+void spacemit_timer_enable(spacemit_timer_t *timer);
+void spacemit_timer_disable(spacemit_timer_t *timer);
+uint64_t spacemit_timer_get_time(spacemit_timer_t *timer);
+void spacemit_timer_reset(spacemit_timer_t *timer);
+int spacemit_timer_set_timeout(spacemit_timer_t *timer, uint64_t ns, bool is_periodic);
+void spacemit_timer_disable_all(void *vaddr);
+void spacemit_timer_init(spacemit_timer_t *timer, void *vaddr, uint64_t channel);
--- a/libplatsupport/src/mach/spacemit/ltimer.c
+++ b/libplatsupport/src/mach/spacemit/ltimer.c
@@ -1,0 +1,256 @@
+/*
+ * Copyright 2025, 10xEngineers
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include <platsupport/io.h>
+#include <platsupport/ltimer.h>
+#include <platsupport/mach/timer.h>
+#include <platsupport/pmem.h>
+#include <platsupport/timer.h>
+#include <utils/io.h>
+#include <utils/util.h>
+
+#include "../../ltimer.h"
+
+enum {
+    COUNTER_TIMER,
+    TIMEOUT_TIMER,
+    NUM_TIMERS
+};
+
+typedef struct {
+    void *vaddr;
+    spacemit_timer_t timer[NUM_TIMERS];
+    irq_id_t irq_id[NUM_TIMERS];
+    timer_callback_data_t callback_data[NUM_TIMERS];
+    ltimer_callback_fn_t user_callback;
+    void *user_callback_token;
+    ps_io_ops_t ops;
+} spacemit_ltimer_t;
+
+static ps_irq_t irqs[] = {
+    {
+        .type = PS_INTERRUPT,
+        .trigger.number = SPACEMIT_TIMER0_IRQ,
+    },
+    {
+        .type = PS_INTERRUPT,
+        .trigger.number = SPACEMIT_TIMER1_IRQ,
+    },
+};
+
+static pmem_region_t pmems[] = {
+    {
+        .type = PMEM_TYPE_DEVICE,
+        .base_addr = SPACEMIT_TIMER_BASE,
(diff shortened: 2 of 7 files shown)


Indanz:
I can have a look if you can provide a datasheet that describes the timer hardware.

It's not an sel4test issue, but an problem with your driver, so it should stay here till proven otherwise.

libplatsupport/mach_include/spacemit/platsupport/mach/timer.h
```diff
@@ -0,0 +1,54 @@
...
+#define SPACEMIT_TIMER_BASE 0xd4014000
+#define SPACEMIT_TIMER_SIZE 0xc8
+#define SPACEMIT_TIMER_IRQ 0x17
+
+/* This is 12.8MHz */
+#define SPACEMIT_TIMER_TICKS_PER_SECOND 0xc35000
+
+#define SPACEMIT_TIMER_MAX_TICKS UINT64_MAX
+
+#define SPACEMIT_NUM_TIMERS 1
```
Indanz: This one is unused.

libplatsupport/src/mach/spacemit/timer.c
```diff
@@ -0,0 +1,141 @@
...
+
+    return value_ns;
+}
+
+void spacemit_timer_reset(spacemit_timer_t *timer)
+{
+    assert(timer);
+    assert(timer->regs);
+
+    timer->regs->ctrl |= SPACEMIT_TIMER_MODE_FREE_RUNNING;
```
Indanz: This doesn't change any bits.

libplatsupport/src/mach/spacemit/timer.c
```diff
@@ -0,0 +1,141 @@
...
+
+    /* Read current counter value */
+    uint32_t hi1, lo, hi2;
+    do {
+        hi1 = timer->regs->cnt_hi;
+        lo = timer->regs->cnt_lo;
+        hi2 = timer->regs->cnt_hi;
+    } while (hi1 != hi2);
+
+    uint64_t current_ticks = (((uint64_t)hi1) << 32) | lo;
```
Indanz: Better to create a small helper function that does this and use it both here and in `spacemit_timer_get_time()`.

libplatsupport/src/mach/spacemit/timer.c
```diff
@@ -0,0 +1,141 @@
...
+    return 0;
+}
+
+void spacemit_timer_disable_all(void *vaddr)
+{
+    assert(vaddr);
+
+    volatile spacemit_timer_regs_t *regs = vaddr;
+    regs->ctrl &= 0xfffffffe;
+}
```
Indanz: Why do you need this?

Indanz:
> > Where did you got the register definitions from?
> 
> not sure but I could not find any for now.  [..] but structurally I followed the hifive-p550 style while writing.  

So you don't know the register layout and can't write a timer driver. There is no point in trying to write one if you don't know what registers to read and what they mean!

> > But were your driver functions called at all?
> 
> Yes, I believe, btw let me drop the logs from u-boot when I load the sel4 image .. I've added multiple prints to find the issue...

Well, at least you wired it correctly to the util_libs driver infrastructure, so the problem is not on the seL4 side of things.

Ivan-Velickovic:
The closest thing to a technical reference manual for the Spacemit K1 SoC I can find is [here](https://developer.spacemit.com/documentation?token=Yk8HwtGbFisyFSkaiBsc1SmCnKg).

Unfortunately I can't see a description of the peripheral timers anywhere.

It seems the only source of truth would be the [Linux driver in their fork](https://github.com/spacemit-com/linux-k1x/blob/k1/drivers/clocksource/timer-k1x.c).

Indanz:
Me too. I found the documents, but not the right timer info.

So at first guess, what's seems missing is configuration of the Timer Clock Control Register?

Indanz:
The timer you need to use is probably:

https://developer.spacemit.com/documentation?token=T7TnwVZz1iPBk1kKwAPc6lyKnNb#11.2-timer-%26-watchdog

and doesn't match the register layout of the timer in this PR.

libplatsupport/mach_include/spacemit/platsupport/mach/timer.h
```diff
@@ -0,0 +1,89 @@
+/*
+ * Copyright 2025, 10xEngineers 
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
```
Indanz: Please add a comment here that explains which timers you are using. This is especially important because the timers have vague names (TIMER1 etc.), and each timer peripheral consists of three timers/counters.

Normally this would also be the place to mention the DTS compat string, but that's unknown for this timer. Its probably an existing timer peripheral used before in other SoCs, we just don't know which one.

libplatsupport/mach_include/spacemit/platsupport/mach/timer.h
```diff
@@ -0,0 +1,89 @@
...
+#pragma once
+
+#include <stdio.h>
+#include <stdint.h>
+#include <stdbool.h>
+
+#define SPACEMIT_TIMER_BASE      0xD4014000
+#define SPACEMIT_TIMER_SIZE      0x2000u
+#define SPACEMIT_TIMER0_IRQ      0x17u
+#define SPACEMIT_TIMER1_IRQ      0x1Au
```
Indanz: I generally follow the datasheet notation, as that makes it easier to check.

0x17 = 23 = timer0_1_irq
0x1A = 26 = timer1_1_irq

Don't you want to use the same peripheral and use timer0_2_irq (24) instead?

akifejaz:
Not sure why [Style](https://github.com/seL4/util_libs/actions/runs/19287460032/job/55176693217?pr=202) is failing. Can someone please help me with this?

libplatsupport/src/mach/spacemit/ltimer.c
```diff
@@ -0,0 +1,252 @@
...
+    long irq_number = irq->irq.number;
+    ltimer_event_t event;
+
+    if (irq_number != irqs[0].irq.number) {
+        ZF_LOGE("Invalid IRQ number %ld received.", irq_number);
+        return EINVAL;
+    }
+
+    /* check Timer Status reg */
+    uint8_t int_status = *(volatile uint32_t *)((uint8_t *)timers->timer[COUNTER_TIMER].vaddr + K1_TSR_OFS_CAL(COUNTER_TIMER));
```
Indanz: Okay this is somewhat hideous. Using offset works great for debugging, but it appears to make the code quite horrible. Considering you have a working driver and already have `spacemit_timer_regs_t`, I propose you add the missing fields and use the structure access instead of the macro and casting stuff. No need to have both `vaddr` and `regs`, `regs` should be enough.

Then you can do `uint32_t int_status = timers->regs->tsr[COUNTER_TIMER];` directly.

akifejaz: Adding all regs in struct `spacemit_timer_regs_t` will make its size 0x100byte, so I thought to use pointers to save some memory. (that is not the constraint but still).

I'll update to structs, that is easier to use, I agree.

libplatsupport/src/mach/spacemit/ltimer.c
```diff
@@ -0,0 +1,252 @@
...
+    if (int_status & 0x1) {
+        spacemit_timer_handle_irq(&timers->timer[COUNTER_TIMER]);
+        event = LTIMER_OVERFLOW_EVENT;
+    }
```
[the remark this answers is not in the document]
akifejaz: I see, I think this would miss `TIMEOUT_TIMER` because that is using `0x2/TMR_Tn_M1` the `COUNTER_TIMER` will work anyway because it is using `TMR_Tn_M0`. 

I need to update the TIMEOUT_TIMER, will do.

libplatsupport/src/mach/spacemit/timer.c
```diff
@@ -0,0 +1,183 @@
...
+{
+    assert(timer);
+	assert(timer->regs);
+
+	uint8_t n = timer->timer_n;
+	timer->value_h += 1;
+
+	/* Clear the interrupt */
+	uint32_t *int_clear = (uint32_t *)((uint8_t *)timer->vaddr + K1_TICLR_OFS_CAL(n));
+	*(int_clear) |= (1u << n);    // using match register as timer number (E.g. Timer0 uses Match0, Timer1 uses Match1, etc)
```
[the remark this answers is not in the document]
akifejaz: but its it in the right place? like this is the interrupt flow i believe is `ltimer_handle_irq` -> `spacemit_timer_handle_irq`

libplatsupport/src/mach/spacemit/ltimer.c
Indanz: It's a description of the memory as it is, you don't allocate it, so its size doesn't matter, as long as it matches the peripheral .

libplatsupport/mach_include/spacemit/platsupport/mach/timer.h
```diff
@@ -25,50 +33,23 @@
...
+    uint32_t ticr[4];       // Timer Interrupt Clear Register (TICLRx)
+    uint32_t tsr[4];        // Timer Status Register (TSRx)
 
-typedef struct {
-    uint32_t tmr_cnt_en;
-    uint32_t tmr_mode;
-    uint32_t tmr_restart;
-    uint32_t tmr_clock_ctrl;
+    uint32_t tcnt[4];       // Timer Count Register (TCRx)
 } spacemit_timer_regs_t;
```
Indanz: It's good to add a static _assert to check that the last field has the correct offset, to catch most errors elsewhere.

libplatsupport/mach_include/spacemit/platsupport/mach/timer.h
```diff
@@ -0,0 +1,68 @@
...
+
+    uint32_t tcnt[4];       // Timer Count Register (TCRx)
+} spacemit_timer_regs_t;
+
+typedef struct {
+    volatile spacemit_timer_regs_t *regs;
+    uint8_t  timer_n;
+    uint32_t value_h;
+    uint32_t *vaddr;
+} spacemit_timer_t;
```
Indanz: To make things nicer, I would get rid of this structure and move `value_h` into `spacemit_ltimer_t` instead, it's only used by the counter. And then pass timer_n as an argument to the common functions.

libplatsupport/src/mach/spacemit/timer.c
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	uint64_t value_ticks = (value_h << 32) | value_l;
+
+	/* convert from ticks to nanoseconds */
+	uint64_t value_whole_seconds = value_ticks / SPACEMIT_TIMER_TICKS_PER_SECOND;
+	uint64_t value_subsecond_ticks = value_ticks % SPACEMIT_TIMER_TICKS_PER_SECOND;
+	uint64_t value_subsecond_ns =
+		(value_subsecond_ticks * NS_IN_S) / SPACEMIT_TIMER_TICKS_PER_SECOND;
+	uint64_t value_ns = value_whole_seconds * NS_IN_S + value_subsecond_ns;
```
Indanz:
```suggestion
	uint64_t whole_seconds = value_ticks / SPACEMIT_TIMER_TICKS_PER_SECOND;
	uint64_t subsecond_ticks = value_ticks % SPACEMIT_TIMER_TICKS_PER_SECOND;
	uint64_t subsecond_ns = (subsecond_ticks * NS_IN_S) / SPACEMIT_TIMER_TICKS_PER_SECOND;
	uint64_t value_ns = whole_seconds * NS_IN_S + subsecond_ns;
```
Prefixing everything with "value_" just makes everything harder to read.

Indanz (approved):
Just some nitpicks, nothing important.

Let me know when you're happy with it. Don't forget to redo a final sel4test run to catch last second mistakes.

akifejaz:
> Just some nitpicks, nothing important.
> 
> Let me know when you're happy with it. Don't forget to redo a final sel4test run to catch last second mistakes.

Thanks for the review, PR is ready from myside. 

btw I did not remove the `spacemit_timer_t` i think it's fine this way too.

akifejaz:
This PR is ready now. 

let me know if needed any change from myside, Thanks!

PR #205, opened 2026-02-12
libutils: Add support for stm32mp2 SoC family

bruelc:
### Description

Add support for uart and timers for stm32mp2 soc

Link to kernel PR [seL4](https://github.com/seL4/seL4/pull/1592)

### Testing

Testing was part of the seL4-test and seL4-bench test suites

### DIFF
--- a/libplatsupport/plat_include/stm32mp2/platsupport/plat/serial.h
+++ b/libplatsupport/plat_include/stm32mp2/platsupport/plat/serial.h
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2026, STMicroelectronics
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#pragma once
+
+#define USART1_PADDR 0x40330000
+#define USART2_PADDR 0x400e0000
+#define USART3_PADDR 0x400f0000
+#define USART6_PADDR 0x40220000
+#define UART4_PADDR  0x40100000
+#define UART5_PADDR  0x40110000
+#define UART7_PADDR  0x40370000
+#define UART8_PADDR  0x40380000
+#define UART9_PADDR  0x402c0000
+
+/* official device names */
+enum chardev_id {
+    USART1,
+    USART2,
+    USART3,
+    USART6,
+    UART4,
+    UART5,
+    UART7,
+    UART8,
+    UART9,
+    /* defaults */
+    PS_SERIAL_DEFAULT = USART2
+};
--- a/libplatsupport/plat_include/stm32mp2/platsupport/plat/timer.h
+++ b/libplatsupport/plat_include/stm32mp2/platsupport/plat/timer.h
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2026, STMicroelectronics
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+#pragma once
+
+#include <platsupport/ltimer.h>
+#include <platsupport/timer.h>
+
+/* Input clock */
+#define TIM2_HZ 200000000UL
+#define TICK_NS 5000UL
+
+/* RCC TIM2 basic timer */
+#define RCC_PADDR      0x44200000
+#define RCC_TIM2_OFF   0x704
+#define RCC_TIM2_PADDR (RCC_PADDR + RCC_TIM2_OFF)
+#define RCC_TIM2_SIZE  4
+#define TIM2_IRQ       137
+#define TIM2_MAP       0x40000000
+
+typedef struct stm32_regs {
+    uint32_t cr1;
+    uint32_t cr2;
+    uint32_t smcr;
+    uint32_t dier;
+    uint32_t sr;
+    uint32_t egr;
+    uint32_t ccmr1;
+    uint32_t ccmr2;
+    uint32_t ccer;
+    uint32_t cnt;
+    uint32_t psc;
+    uint32_t arr;
+    uint32_t rcr; /* n/a for TIM2 */
+    uint32_t ccr1;
+    uint32_t ccr2;
+    uint32_t ccr3;
+    uint32_t ccr4;
+} stm32_regs_t;
+
+#define STM32_TIM_CCER_CC1E   BIT(0)
+
+#define STM32_TIM_CR1_CEN     BIT(0)
+#define STM32_TIM_CR1_UDIS    BIT(1)
+#define STM32_TIM_CR1_URS     BIT(1)
+#define STM32_TIM_CR1_OPM     BIT(6)
+#define STM32_TIM_CR1_ARPE    BIT(7)
+
+#define STM32_TIM_EGR_UG      BIT(0)
+
+#define STM32_TIM_DIER_UIE    BIT(0)
+#define STM32_TIM_DIER_CC1IE  BIT(1)
+
+#define STM32_TIM_SR_UIF      BIT(0)
+#define STM32_TIM_SR_CC1IF    BIT(1)
+
+typedef struct {
+    volatile stm32_regs_t *hw;
+    uint32_t periodic_t;
+    uint32_t cnt_h; /* keep overflow for get_time */
+} stm32_t;
+
+int stm32mp2_timer_init(stm32_t *stm32, ps_io_ops_t ops);
+void stm32_timer_reset(stm32_t *stm32);
+uint64_t stm32_get_time(stm32_t *stm32);
+int stm32_set_timeout(stm32_t *dmt, uint64_t ns, bool periodic);
+int stm32_start_timer(stm32_t *stm32);
+int stm32_stop_timer(stm32_t *stm32);
--- a/libplatsupport/src/plat/stm32mp2/chardev.c
+++ b/libplatsupport/src/plat/stm32mp2/chardev.c
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2026, STMicroelectronics
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include "../../chardev.h"
+#include "../../common.h"
+#include <utils/util.h>
+
+#define UART_DEFN(devid) {         \
+    .id      = UART##devid,         \
+    .paddr   = UART##devid##_PADDR, \
+    .size    = BIT(12),             \
+    .init_fn = &uart_init           \
+}
+
+#define USART_DEFN(devid) {         \
+    .id      = USART##devid,         \
+    .paddr   = USART##devid##_PADDR, \
+    .size    = BIT(12),             \
+    .init_fn = &uart_init           \
+}
+
+static const struct dev_defn dev_defn[] = {
+    USART_DEFN(1),
+    USART_DEFN(2),
+    USART_DEFN(3),
+    USART_DEFN(6),
+    UART_DEFN(4),
+    UART_DEFN(5),
+    UART_DEFN(7),
+    UART_DEFN(8),
+    UART_DEFN(9),
+};
+
+struct ps_chardevice *
+ps_cdev_init(enum chardev_id id, const ps_io_ops_t *o, struct ps_chardevice *d)
+{
+    unsigned int i;
+    for (i = 0; i < ARRAY_SIZE(dev_defn); i++) {
+        if (dev_defn[i].id == id) {
+            return (dev_defn[i].init_fn(dev_defn + i, o, d)) ? NULL : d;
+        }
+    }
+    return NULL;
+}
--- a/libplatsupport/src/plat/stm32mp2/ltimer.c
+++ b/libplatsupport/src/plat/stm32mp2/ltimer.c
@@ -1,0 +1,186 @@
+/*
+ * Copyright 2026, STMicroelectronics
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+#include <stdio.h>
+
+#include <utils/util.h>
+#include <utils/time.h>
+
+#include <platsupport/plat/timer.h>
+#include <platsupport/io.h>
+
+#include "../../ltimer.h"
+
+typedef struct {
+    stm32_t stm32_timer;
+    irq_id_t irq_id;
+    timer_callback_data_t callback_data;
+    ltimer_callback_fn_t user_cb_fn;
+    void *user_cb_token;
+    ps_io_ops_t ops;
+} stm32_ltimer_t;
+
+static pmem_region_t pmems = {
+    .type = PMEM_TYPE_DEVICE,
+    .base_addr = TIM2_MAP,
+    .length = PAGE_SIZE_4K
+};
+
+static ps_irq_t irqs = {
+    .type = PS_INTERRUPT,
+    .irq.number = TIM2_IRQ,
+};
+
+static int get_time(void *data, uint64_t *time)
+{
+    stm32_ltimer_t *stm32_ltimer = data;
+
+    *time = stm32_get_time(&stm32_ltimer->stm32_timer);
+
+    return 0;
+}
+
+static int set_timeout(void *data, uint64_t ns, timeout_type_t type)
+{
+    uint64_t timeout;
+    stm32_ltimer_t *stm32_ltimer = data;
+    uint64_t time = stm32_get_time(&stm32_ltimer->stm32_timer);
+
+    switch (type) {
+    case TIMEOUT_ABSOLUTE:
+        if ((ns - time) / TICK_NS > UINT32_MAX) {
+            ZF_LOGE("Timeout too far in the future");
+            return ETIME;
+        }
+        timeout = ns;
+        break;
+    case TIMEOUT_RELATIVE:
+    case TIMEOUT_PERIODIC:
+        if (ns / TICK_NS > UINT32_MAX) {
+            ZF_LOGE("Timeout too far in the future");
+            return ETIME;
+        }
+        timeout = ns + time;
+        break;
+    }
+
+    return stm32_set_timeout(&stm32_ltimer->stm32_timer, timeout, type == TIMEOUT_PERIODIC);
+}
+
+static int reset(void *data)
+{
+    stm32_ltimer_t *stm32_ltimer = data;
+
+    stm32_stop_timer(&stm32_ltimer->stm32_timer);
+    stm32_start_timer(&stm32_ltimer->stm32_timer);
+
+    return 0;
+}
+
+static void destroy(void *data)
+{
+    stm32_ltimer_t *stm32_ltimer = data;
+
+    stm32_stop_timer(&stm32_ltimer->stm32_timer);
+
+    if (stm32_ltimer->stm32_timer.hw) {
+        ps_pmem_unmap(&stm32_ltimer->ops, pmems, (void *) stm32_ltimer->stm32_timer.hw);
+    }
+
+    if (stm32_ltimer->irq_id != PS_INVALID_IRQ_ID) {
+        int error = ps_irq_unregister(&stm32_ltimer->ops.irq_ops, stm32_ltimer->irq_id);
+        ZF_LOGF_IF(error, "Failed to unregister IRQ");
+    }
+
+    ps_free(&stm32_ltimer->ops.malloc_ops, sizeof(stm32_ltimer_t), stm32_ltimer);
+}
+
+static int handle_irq(void *data, ps_irq_t *irq)
+{
+    stm32_ltimer_t *stm32_ltimer = data;
+
+    long irq_number = irq->irq.number;
+    stm32_t *stm32 = &stm32_ltimer->stm32_timer;
+    volatile stm32_regs_t *stm32_regs = stm32->hw;
+    uint32_t sr;
+
+    assert(irq_number == irqs.irq.number);
+
+    sr = stm32_regs->sr;
+
+    if (sr & STM32_TIM_SR_UIF) {
+        stm32->cnt_h++;
+        sr &= ~STM32_TIM_SR_UIF;
+
+        if (stm32_ltimer->user_cb_fn) {
+            stm32_ltimer->user_cb_fn(stm32_ltimer->user_cb_token, LTIMER_OVERFLOW_EVENT);
+        }
+    }
+
+    if (sr & STM32_TIM_SR_CC1IF) {
+        if (stm32->periodic_t) {
+            stm32->hw->ccr1 += stm32->periodic_t;
+        } else {
+            stm32_regs->dier &= ~STM32_TIM_DIER_CC1IE;
+        }
+        sr &= ~STM32_TIM_SR_CC1IF;
+
+        if (stm32_ltimer->user_cb_fn) {
+            stm32_ltimer->user_cb_fn(stm32_ltimer->user_cb_token, LTIMER_TIMEOUT_EVENT);
+        }
+    }
+
+    stm32_regs->sr = sr;
+
+    return 0;
+}
+
+int ltimer_default_init(ltimer_t *ltimer, ps_io_ops_t ops, ltimer_callback_fn_t callback, void *callback_token)
+{
+    int rc;
+    void *stm32_map_base;
+
+    rc = create_ltimer_simple(ltimer, ops, sizeof(stm32_ltimer_t), get_time,
+                              set_timeout, reset, destroy);
+    if (rc) {
+        ZF_LOGE("ltimer creation failed");
+        return rc;
+    }
+
+    stm32_ltimer_t *stm32_ltimer = ltimer->data;
+    stm32_ltimer->ops = ops;
+
+    stm32_t *stm32 = &stm32_ltimer->stm32_timer;
+
+    stm32_ltimer->user_cb_fn = callback;
+    stm32_ltimer->user_cb_token = callback_token;
+
+    stm32_map_base = ps_pmem_map(&ops, pmems, false, PS_MEM_NORMAL);
+    if (stm32_map_base == NULL) {
+        destroy(ltimer->data);
+        return EINVAL;
+    }
+    stm32_ltimer->stm32_timer.hw = stm32_map_base;
+
+    stm32_ltimer->callback_data.ltimer = ltimer;
+    stm32_ltimer->callback_data.irq_handler = handle_irq;
+    stm32_ltimer->callback_data.irq = &irqs;
+
+    stm32_ltimer->irq_id = ps_irq_register(&ops.irq_ops, irqs, handle_irq_wrapper,
+                                           &stm32_ltimer->callback_data);
+    if (stm32_ltimer->irq_id < 0) {
+        destroy(ltimer->data);
+        return EIO;
+    }
+
+    rc = stm32mp2_timer_init(&stm32_ltimer->stm32_timer, ops);
+    if (rc) {
+        ZF_LOGE("Failed to init stm32 timeout timer");
+        destroy(&stm32_ltimer);
+        return rc;
+    }
+
+    return 0;
+}
--- a/libplatsupport/src/plat/stm32mp2/timer.c
+++ b/libplatsupport/src/plat/stm32mp2/timer.c
@@ -1,0 +1,114 @@
+/*
+ * Copyright 2026, STMicroelectronics
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+/* Use 32 bit TIM2 general timer to handle 64 bit timestamp with ARR
+   overflow in up-counting mode. CNT is reset at earch overflow.
+   32 bit timeout interrupt is supported with Channel 1 output compare. */
+
+#include <stdio.h>
+#include <errno.h>
+#include <platsupport/timer.h>
+#include <platsupport/plat/timer.h>
+#include <utils/frequency.h>
+
+#include "../../services.h"
+#include "../../ltimer.h"
+
+static void print_regs(stm32_t *stm32)
+{
+    printf("control register 1  >> 0x%08x\n", stm32->hw->cr1);
+    printf("control register 2  >> 0x%08x\n", stm32->hw->cr2);
+    printf("dier                >> 0x%08x\n", stm32->hw->dier);
+    printf("status register     >> 0x%08x\n", stm32->hw->sr);
+    printf("egr                 >> 0x%08x\n", stm32->hw->egr);
+    printf("ccr1                >> 0x%08x\n", stm32->hw->ccr1);
+    printf("counter             >> 0x%08x\n", stm32->hw->cnt);
+    printf("prescaler           >> 0x%08x\n", stm32->hw->psc);
+    printf("arr                 >> 0x%08x\n", stm32->hw->arr);
+}
+
+int stm32_start_timer(stm32_t *stm32)
+{
+    stm32->hw->cnt = 0;
+    stm32->cnt_h = 0;
+
+    stm32->hw->cr1 |= STM32_TIM_CR1_CEN;
+
+    return 0;
+}
+
+int stm32_stop_timer(stm32_t *stm32)
+{
+    stm32->hw->cr1 &= ~STM32_TIM_CR1_CEN;
+    stm32->hw->dier &= ~STM32_TIM_DIER_CC1IE;
+
+    return 0;
+}
+
+int stm32_set_timeout(stm32_t *stm32, uint64_t timeout, bool periodic)
+{
+    uint32_t target = timeout / TICK_NS;
+
+    stm32->hw->ccr1 = target;
+    stm32->hw->sr &= ~STM32_TIM_SR_CC1IF;
+    stm32->hw->dier |= STM32_TIM_DIER_CC1IE;
+    stm32->periodic_t = periodic ? target : 0UL;
+
+    return 0;
+}
+
+uint64_t stm32_get_time(stm32_t *stm32)
+{
+    uint32_t cnt = stm32->hw->cnt;
+    uint32_t hi = stm32->cnt_h;
+
+    /* Check for pending overflow. */
+    if (stm32->hw->sr & STM32_TIM_SR_UIF) {
+        hi++;
+        /* In case it happened after the last read */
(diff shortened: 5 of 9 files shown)


libplatsupport/plat_include/stm32mp2/platsupport/plat/serial.h
```diff
@@ -0,0 +1,22 @@
...
+#define UART1_PADDR  0x400F0000
+
+#define DEFAULT_SERIAL_PADDR     UART0_PADDR
+#define DEFAULT_SERIAL_INTERRUPT UART0_IRQ
+#define UART_REF_CLK             64000000
+
+/* official device names */
+enum chardev_id {
+    UART0,
+    UART1,
```
Indanz: Is UART1 actually supported by this driver? If not, leave it out.

The SoC has about 10 UARTs, probably a good idea to follow the datasheet UART names in this enum and elsewhere.

libplatsupport/plat_include/stm32mp2/platsupport/plat/serial.h
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright 2026, STMicroelectronics
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#pragma once
+
+#define UART0_PADDR  0x400E0000
```
Indanz: This is actually USART2.

libplatsupport/plat_include/stm32mp2/platsupport/plat/serial.h
```diff
@@ -0,0 +1,22 @@
...
+#define UART1_PADDR  0x400F0000
```
Indanz: And this is USART3.

libplatsupport/plat_include/stm32mp2/platsupport/plat/timer.h
```diff
@@ -0,0 +1,99 @@
+/*
+ * Copyright 2026, STMicroelectronics
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
```
Indanz, line 5: It's good to give an overview here of which timers you're using and how and why (if it's more than one).

libplatsupport/plat_include/stm32mp2/platsupport/plat/timer.h
```diff
@@ -0,0 +1,99 @@
...
+#define RCC_TIM_SIZE       8
+
+#define STM32_TIM2_PATH "/soc@0/bus@42080000/timer@40000000"
+#define STM32_TIM3_PATH "/soc@0/bus@42080000/timer@40010000"
+
+#define RCC_ON(rcc)    (*(volatile uint32_t *)(rcc) = 6)
+
+#define RCC_APB1DIVR_OFF  0x4B4
+#define RCC_APB1DIVR_PADDR      (RCC_PADDR + RCC_APB1DIVR_OFF)
+#define RCC_APB1DIVR_SIZE       4
```
Indanz: Weird indentation. And also unused?

libplatsupport/src/plat/stm32mp2/chardev.c
```diff
@@ -0,0 +1,35 @@
...
+ * Copyright 2026, STMicroelectronics
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include "../../chardev.h"
+#include "../../common.h"
+#include <utils/util.h>
+
+#include "../../chardev.h"
```
Indanz: Duplicate, and see Julia's comment elsewhere.

libplatsupport/plat_include/stm32mp2/platsupport/plat/timer.h
```diff
@@ -0,0 +1,99 @@
...
+    uint32_t ccr2;
+    uint32_t ccr3;
+    uint32_t ccr4;
+    uint32_t bdtr;
+    uint32_t dcr;
+    uint32_t dmar;
+    uint32_t tisel;
+    uint32_t pad[0xEA];
+    uint32_t ipidr;
+} stm32_regs_t;
```
Indanz, line 41: This structure you can move to timer.c instead, leaving a forward declaration behind. Same for most of the defines. But whatever you prefer.

libplatsupport/src/plat/stm32mp2/chardev.c
```diff
@@ -0,0 +1,35 @@
...
+    .id      = UART##devid,         \
+    .paddr   = UART##devid##_PADDR, \
+    .size    = BIT(12),             \
+    .init_fn = &uart_init           \
+}
+
+static const struct dev_defn dev_defn[] = {
+    UART_DEFN(0),
+    UART_DEFN(1),
+};
```
Indanz, line 35: If you use this construction instead of hard-coding a specific UART, any reason not to add all compatible ones?

libplatsupport/src/plat/stm32mp2/ltimer.c
```diff
@@ -0,0 +1,151 @@
...
+
+#include "../../services.h"
+
+#include <platsupport/plat/timer.h>
+#include <platsupport/io.h>
+
+#include "../../ltimer.h"
+
+/*
+ * We use two dm timers: one to keep track of an absolute time, the other for timeouts.
```
Indanz: You're using the TIM2 and TIM3 general purpose timers. What do you mean with "dm"?

Also, each timer seems to have four channels, so why do you need to use two timers?

Indanz (commented):
You don't handle 32-bit timer overflows, you probably want to register an IRQ handler for that. Alternatively, you can probably link two timers into one 64-bit timer and ignore overflows.

You also use two hard-coded timers, instead of one. Each timer has four channels which can be used for compare independently. I'd use one for the timeout function and another for overflow detection, if there is no other way. Except if it's one of those stupid timers which can't be reprogrammed on-the-fly without stopping it, in which case you do need to use two timers and document this somewhere. But the datasheet says _"The TIMx_CCRx register can be updated at any time by software to control the output waveform, provided that the preload register is not enable"_, so it seems okay.

But if hard-coding timers like this, it doesn't make much sense to use the DTS walker either, as you know everything already. (Keep it if the code is simpler though.)

libplatsupport/plat_include/stm32mp2/platsupport/plat/serial.h
bruelc: Correct. In the boot stdout chosen dts path, we have serial0 aliased to USART2, which confused my naming. I will rename it.

libplatsupport/src/plat/stm32mp2/ltimer.c
bruelc: No specific reason except that it was easier to debug and trace. But OK, now I should rework this part ; there is no need to start more than necessary.

bruelc:
Indanz, your suggestion to use a channel instead of a second timer was very good. I force-pushed a new version with your latest reviews and refactoring. Thank you.

libplatsupport/src/plat/stm32mp2/chardev.c
```diff
@@ -0,0 +1,32 @@
...
+
+#define UART_DEFN(devid) {         \
+    .id      = USART##devid,         \
+    .paddr   = USART##devid##_PADDR, \
+    .size    = BIT(12),             \
+    .init_fn = &uart_init           \
+}
+
+static const struct dev_defn dev_defn[] = {
+    UART_DEFN(1),
```
Indanz: I'd like to see the other compatible UARTS in this list too, if possible. And if the address is calculable, do that instead of macro string concatenation.

If the only difference is the memory address then there is no reason not to add them. No need to test the others in that case either. Even if it doesn't work because of e.g. missing clock or power domain enable, it's still better than nothing.

libplatsupport/src/plat/stm32mp2/ltimer.c
```diff
@@ -0,0 +1,195 @@
...
+		    stm32_regs->dier &= ~STM32_TIM_DIER_CC1IE;
+	    }
+            else
+              assert(0);
+
+            if (stm32_ltimer->user_cb_fn) {
+                stm32_ltimer->user_cb_fn(stm32_ltimer->user_cb_token, LTIMER_TIMEOUT_EVENT);
+            }
+
+    }
```
Indanz: Please cleanup this bit, looks like left-over debug stuff.

libplatsupport/src/plat/stm32mp2/ltimer.c
```diff
@@ -0,0 +1,195 @@
...
+        return EINVAL;
+    }
+
+    if ((stm32_regs->sr & STM32_TIM_SR_UIF) && (stm32_regs->dier & STM32_TIM_DIER_UIE)) {
+	    stm32_regs->sr &= ~STM32_TIM_SR_UIF;
+	    stm32->cnt_h++;
+
+            if (stm32_ltimer->user_cb_fn) {
+                stm32_ltimer->user_cb_fn(stm32_ltimer->user_cb_token, LTIMER_OVERFLOW_EVENT);
+            }
```
Indanz: Indent off.

libplatsupport/src/plat/stm32mp2/ltimer.c
```diff
@@ -0,0 +1,195 @@
...
+
+int ltimer_default_init(ltimer_t *ltimer, ps_io_ops_t ops, ltimer_callback_fn_t callback, void *callback_token)
+{
+    int rc;
+    void *stm32_map_base;
+
+    if (ltimer == NULL) {
+        ZF_LOGE("ltimer cannot be NULL");
+        return EINVAL;
+    }
```
Indanz: Is this check really necessary? Seems a bit pointless to me, something is going to go horribly wrong sooner or later if it is NULL.

libplatsupport/src/plat/stm32mp2/timer.c
```diff
@@ -0,0 +1,134 @@
...
+    printf("ccr1                >> 0x%08x\n", stm32->hw->ccr1);
+    printf("counter             >> 0x%08x\n", stm32->hw->cnt);
+    printf("prescaler           >> 0x%08x\n", stm32->hw->psc);
+    printf("arr                 >> 0x%08x\n", stm32->hw->arr);
+}
+
+int stm32_start_timer(stm32_t *stm32)
+{
+    assert(stm32 != NULL);
+    assert(stm32->hw  != NULL);
```
Indanz: All these asserts can go away once your driver is working, it's only useful to catch initial problems.

libplatsupport/src/plat/stm32mp2/timer.c
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    MAP_IF_NULL(io_ops, RCC_TIM2, rcc);
+    if (rcc == NULL)
+        return EINVAL;
+
+    /* TIM2 RCC */
+    RCC_ON(rcc);
+
+    stm32->hw->cr1 &= ~STM32_TIM_CR1_CEN;
```
Indanz: I think you should do blind writes to cr1 to configure it how you want in one go. Currently you never set the other bits, so you're counting on either reset behaviour or something.

Reading registers is very slow, so if you don't have to, don't (this applies more to the other places than here). Some devices don't want other bits to be changed together with the enable bit, but a lot of them don't mind.

libplatsupport/src/plat/stm32mp2/timer.c
```diff
@@ -0,0 +1,134 @@
...
+
+    stm32->hw->psc = (TICK_NS * TIM2_MHZ / NS_IN_S) - 1;
+    stm32->hw->arr = UINT32_MAX;
+    stm32->hw->sr = 0;
+    stm32->hw->dier |= STM32_TIM_DIER_UIE; /* interrupt for overflow */
+    stm32->hw->ccmr1 = 0;
+    stm32->hw->ccer |= STM32_TIM_CCER_CC1E;
+    stm32->hw->dier &= ~STM32_TIM_DIER_CC1IE;
```
Indanz: Same for the `ccer` and `dier` registers, I think you just want to set it to `STM32_TIM_CCER_CC1E` and `STM32_TIM_DIER_UIE`. Or are there reserved bits in there?

libplatsupport/src/plat/stm32mp2/ltimer.c
bruelc: > Please cleanup this bit, looks like left-over debug stuff.

Gloops wrong version pushed. Please forget this commit

bruelc: repushed (with pediodic support). now I'll look at your other comments :)

libplatsupport/src/plat/stm32mp2/chardev.c
bruelc: humm, no real reason, all the potential platforms use the same default. But ok I can add the other ones in case

libplatsupport/src/plat/stm32mp2/ltimer.c
bruelc: Yes, I aligned this check with the other platforms. Looking at the call sites, it is indeed not null, but this is a strong assumption.
There is an assert in create_ltimer_simple, so we do have a guard in debug mode.

libplatsupport/src/plat/stm32mp2/timer.c
bruelc: OK . I'm usually suspicious with reserved bits, here there are a bunch of reserved bits that are 0 at reset.  We should be safe to write 0, in this revision at least.

libplatsupport/src/plat/stm32mp2/timer.c
bruelc: there are reserved bits, but 0 at reset and (looking from the HAL code) we can rewrite them

Indanz: I really hate hardware that says that some bits are reserved and that you can't modify them, and at the same time keep them undefined. That forces unnecessary read-modify-writes. Please make sure no one at ST introduces such crappy HW interfaces.

bruelc: > I really hate hardware that says that some bits are reserved and that you can't modify them, and at the same time keep them undefined. That forces unnecessary read-modify-writes. Please make sure no one at ST introduces such crappy HW interfaces.

This is not the case. The HW does not keep these bits undefined. The reference manual states for these bits: "Reserved; must be kept at reset value". The reset value is 0x0000. I think the latest push matches this

Indanz: I know it doesn't apply to this case, but it was a generic comment for hardware that does.

libplatsupport/plat_include/stm32mp2/platsupport/plat/timer.h
```diff
@@ -0,0 +1,77 @@
...
+typedef struct {
+    volatile stm32_regs_t *hw;
+    uint64_t periodic_ns;
+    uint32_t cnt_h; /* keep overflow for get_time */
+} stm32_t;
+
+typedef struct {
+    /* set in init */
+    const char *fdt_path;
+} stm32_config_t;
```
Indanz: This is now unused and can be removed.

libplatsupport/src/plat/stm32mp2/ltimer.c
```diff
@@ -0,0 +1,191 @@
...
+typedef struct {
+    stm32_t stm32_timer;
+    irq_id_t irq_id;
+    timer_callback_data_t callback_data;
+    ltimer_callback_fn_t user_cb_fn;
+    void *user_cb_token;
+    ps_io_ops_t ops;
+} stm32_ltimer_t;
+
+static pmem_region_t pmems[] = {
```
Indanz: You are only using pmems[0], no need to make this an array.

libplatsupport/src/plat/stm32mp2/ltimer.c
```diff
@@ -0,0 +1,191 @@
...
+    }
+};
+
+static ps_irq_t irqs[] = {
+    {
+        .type = PS_TRIGGER,
+        .trigger.number = TIM2_IRQ,
+        .trigger.trigger = 1
+    }
+};
```
Indanz: Same here.

libplatsupport/src/plat/stm32mp2/ltimer.c
```diff
@@ -0,0 +1,191 @@
...
+
+static int set_timeout(void *data, uint64_t ns, timeout_type_t type)
+{
+    uint64_t timeout, time;
+    stm32_ltimer_t *stm32_ltimer = data;
+
+    switch (type) {
+    case TIMEOUT_ABSOLUTE:
+	timeout = ns;
+        break;
```
Indanz, line 58: Indentation is off.

libplatsupport/src/plat/stm32mp2/ltimer.c
```diff
@@ -0,0 +1,191 @@
...
+    case TIMEOUT_RELATIVE:
+    case TIMEOUT_PERIODIC:
+        time = stm32_get_time(&stm32_ltimer->stm32_timer);
+	timeout = ns + time;
+        break;
```
Indanz, line 66: Same.

libplatsupport/src/plat/stm32mp2/ltimer.c
```diff
@@ -0,0 +1,191 @@
...
+        return EINVAL;
+    }
+
+    if ((stm32_regs->sr & STM32_TIM_SR_UIF) && (stm32_regs->dier & STM32_TIM_DIER_UIE)) {
+	    stm32_regs->sr &= ~STM32_TIM_SR_UIF;
+	    stm32->cnt_h++;
+
+            if (stm32_ltimer->user_cb_fn) {
+                stm32_ltimer->user_cb_fn(stm32_ltimer->user_cb_token, LTIMER_OVERFLOW_EVENT);
+            }
```
Indanz:
```suggestion
        if (stm32_ltimer->user_cb_fn) {
            stm32_ltimer->user_cb_fn(stm32_ltimer->user_cb_token, LTIMER_OVERFLOW_EVENT);
        }
```

libplatsupport/src/plat/stm32mp2/ltimer.c
```diff
@@ -0,0 +1,191 @@
...
+              if (((uint32_t)target > UINT32_MAX) ||
+                  (stm32->cnt_h && ((uint32_t)target < stm32->hw->cnt))) {
+                ZF_LOGW("%llu ticks overflow not supported \n", target);
+                return ETIME;
+              }
+	    }
+
+            if (stm32_ltimer->user_cb_fn) {
+                stm32_ltimer->user_cb_fn(stm32_ltimer->user_cb_token, LTIMER_TIMEOUT_EVENT);
+            }
```
Indanz: Indentation:
```suggestion
	    } else {
            uint64_t time = stm32_get_time(&stm32_ltimer->stm32_timer);
            uint64_t target = stm32->periodic_ns + time;
            target = target / TICK_NS;
            if (((uint32_t)target > UINT32_MAX) ||
                (stm32->cnt_h && ((uint32_t)target < stm32->hw->cnt))) {
                ZF_LOGW("%llu ticks overflow not supported \n", target);
                return ETIME;
            }
	    }
        if (stm32_ltimer->user_cb_fn) {
            stm32_ltimer->user_cb_fn(stm32_ltimer->user_cb_token, LTIMER_TIMEOUT_EVENT);
        }
```

libplatsupport/src/plat/stm32mp2/ltimer.c
```diff
@@ -0,0 +1,191 @@
...
+	    if (! stm32->periodic_ns) {
+		    stm32_regs->dier &= ~STM32_TIM_DIER_CC1IE;
+	    }
+            else {
+              uint64_t time = stm32_get_time(&stm32_ltimer->stm32_timer);
+              uint64_t target = stm32->periodic_ns + time;
+              target = target / TICK_NS;
+              if (((uint32_t)target > UINT32_MAX) ||
+                  (stm32->cnt_h && ((uint32_t)target < stm32->hw->cnt))) {
+                ZF_LOGW("%llu ticks overflow not supported \n", target);
```
Indanz: Why not? Should be trivial now you handle overflows.

To word it stronger: You don't want your timer to be unusable just before it wraps around. So the `UINT32_MAX` check is fine, but the rest really isn't.

Indanz (approved):
Probably need to fix the style a bit, but otherwise looks good.

libplatsupport/src/plat/stm32mp2/ltimer.c
bruelc: yeah I don't know what I had in mind.  the ((uint32_t)target < stm32->hw->cnt test is stupid

for a periodic timer if cnt_h is set, target is in the same time domain...

bruelc: and the test: if (((uint32_t)target > UINT32_MAX) is stupid too :)  I don't want the cast...

libplatsupport/src/plat/stm32mp2/timer.c
```diff
@@ -0,0 +1,112 @@
...
+
+int stm32_set_timeout(stm32_t *stm32, uint64_t ns, bool periodic)
+{
+    uint64_t target = ns / TICK_NS;
+    uint32_t ccr1_h;
+
+    ccr1_h = (uint32_t)(target >> 32);
+
+    /* todo: timeout will overflow. */
+    if (ccr1_h != stm32->cnt_h) {
```
Indanz: For timers that do an exact match instead of greater than check, overflowing once should be fine, as long as you have a `UINT32_MAX` check and don't overflow twice.

Because if it's less than the range, in case of overflow the target time will end up in the past and can only be reached after overflowing once.

libplatsupport/src/plat/stm32mp2/timer.c
```diff
@@ -0,0 +1,112 @@
...
+uint64_t stm32_get_time(stm32_t *stm32)
+{
+    uint32_t cnt;
+    uint64_t cnt2;
+    uint32_t high1, high2;
+
+    if (!stm32->hw->cr1) {
+        ZF_LOGE("timestamp not started");
+        return EIO;
+    }
```
Indanz: You really don't want such an expensive register read to catch programmer errors. If you must have this check, then check a variable in memory.

libplatsupport/src/plat/stm32mp2/timer.c
```diff
@@ -0,0 +1,112 @@
...
+
+    do {
+        high1 = stm32->cnt_h;
+        cnt = stm32->hw->cnt;
+        high2 = stm32->cnt_h;
+    } while (high1 != high2);
```
Indanz: This construction only makes sense for annoying timers that are 64-bit, but have split up the value in two 32-bit registers which you can't read atomically. (Some have split 32-bit registers, but if you read them in the right order they give the correct results, because one half is only latched when the other half is read.)

Here `stm32->cnt_h` is a value in memory and the timer task is single threaded, no need for this loop and double checking high.

However, there is a chance that you read `cnt` just after the timer overflowed, and as the interrupt handler is in the same task, it may not have run yet. So you do need a check for pending overflows. Interrupt handlers are not pre-emptive in user space drivers (except if you make it a separate, higher priority thread, but util_libs doesn't do that).

The expensive way is checking whether the overflow IRQ is pending. The cheap way is storing the last 64-bit get_time value and compare against that (and if your timer wraps around quickly enough, also update it in the overflow interrupt handler by setting it to `cnt_h << 32` after updating cnt_h).

libplatsupport/plat_include/stm32mp2/platsupport/plat/timer.h
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+#pragma once
+
+#include <platsupport/ltimer.h>
+#include <platsupport/timer.h>
+
+/* Input clock */
+#define TIM2_MHZ 200000000UL
```
Indanz:
```suggestion
#define TIM2_HZ 200000000UL
```
Except if the clock is really that high, in which case you really want to set a prescaler...

libplatsupport/src/plat/stm32mp2/ltimer.c
```diff
@@ -0,0 +1,185 @@
...
+    stm32_t *stm32 = &stm32_ltimer->stm32_timer;
+    volatile stm32_regs_t *stm32_regs = stm32->hw;
+
+    if (irq_number != irqs.irq.number) {
+        ZF_LOGE("Invalid IRQ number %ld received.", irq_number);
+        return EINVAL;
+    }
+
+    if ((stm32_regs->sr & STM32_TIM_SR_UIF) && (stm32_regs->dier & STM32_TIM_DIER_UIE)) {
+        stm32_regs->sr &= ~STM32_TIM_SR_UIF;
```
Indanz: The fact that you need to clear interrupt flags increases the chance that it are level-triggered IRQs, not edge-triggered. So I suspect you may want to use `PS_INTERRUPT` instead of `PS_TRIGGER`, worth double checking.

libplatsupport/src/plat/stm32mp2/ltimer.c
```diff
@@ -0,0 +1,185 @@
...
+        stm32->cnt_h++;
+
+        if (stm32_ltimer->user_cb_fn) {
+            stm32_ltimer->user_cb_fn(stm32_ltimer->user_cb_token, LTIMER_OVERFLOW_EVENT);
+        }
+    }
+
+    if ((stm32_regs->sr & STM32_TIM_SR_CC1IF) && (stm32_regs->dier & STM32_TIM_DIER_CC1IE)) {
+        stm32_regs->sr &= ~STM32_TIM_SR_CC1IF;
```
Indanz: Just a comment: You do 6 very slow register reads, of which 5 are strictly speaking not necessary.

libplatsupport/src/plat/stm32mp2/ltimer.c
```diff
@@ -0,0 +1,185 @@
...
+        if (! stm32->periodic_ns) {
+            stm32_regs->dier &= ~STM32_TIM_DIER_CC1IE;
+        } else {
+            uint64_t time = stm32_get_time(&stm32_ltimer->stm32_timer);
+            uint64_t target = (stm32->periodic_ns + time) / TICK_NS;
+            uint32_t ccr1_h = (uint32_t)(target >> 32);
+
+            /* cross domain timeout not supported yet */
+            assert (ccr1_h == stm32->cnt_h);
+            stm32->hw->ccr1 = (uint32_t)target;
```
Indanz: What's this bit for? Doesn't the timer support auto-reload for periodic timers?

Doing get_time here will introduce timing drift which is exactly the thing that you don't want in a periodic timer.

Indanz (commented):
Apparently I had a pending review here which I forgot to submit according to Github?

bruelc:
> Apparently I had a pending review here which I forgot to submit according to Github?

Most of the comments are handled in lates my push, let me check

libplatsupport/plat_include/stm32mp2/platsupport/plat/timer.h
bruelc: Oops Good catch 🙂

libplatsupport/src/plat/stm32mp2/ltimer.c
bruelc: Yes, the update and compare events are high level triggered. I see the irq_trigger_type_t enum in irq.h, but I am not sure about the relationship or use of 'irq_type_t'.

For my information, 'irq_trigger_type_t' does not seem to be used. Is it a leftover from older API?

libplatsupport/src/plat/stm32mp2/ltimer.c
bruelc: I removed the DIER check (if we get here, it's because the IRQ was enabled) in the latest version.
However, I think the SR check is still necessary to detect if we have an overflow or timeout, and clearing it is also required.

libplatsupport/src/plat/stm32mp2/ltimer.c
bruelc: we have auto-reload for the ARR register, which is used of overflow events. But not of channel compare events

The problem here is that the compare event is triggered when CNT reaches CCR1, which was initialized relatively to CNT in set_timeout. So now we have a new CNT base value so we recompute the  target.

Maybe I misunderstood the meaning of "periodic" here in the ltimer interface.
Let's say we have a 100t periodic timer, and the CNT was initially at 10t (not counting the gettime conversion for illustration). So, CCR1 is set to 110t in set_timeout.
When it triggers at CNT = 110t, we now need to retarget CCR1 to 210t, plus maybe a few ticks lost in the handler.
Does that seem sensible?

bruelc: Maybe in this case, the only way to avoid this drift, would be to revert to the original use of 2 different timers that provides the 2 auto-reload registers.
Initially I didn't have a good reason to use 2 timers. Now I have :)

bruelc: after rethinking it, we could do it the other way around: use the channel compare CCR1 to catch overflow, and ARR auto-reload for timeout. But this does not help neither since the CNT base change

bruelc: note ARR reload implies a CNT reset, which is used for timekeeping. If we want to avoid the drift I see no other way than using 2 different timers

bruelc: We can also estimate the drift and substract it from the new target . But that implies reading CNT again. I still prefer 2 timers

bruelc: testing a fix for the drift problem by just reprogramming ccr1 = ccr1 + periodic_ticks. In fact we don't need to auto-reload anything, and overflowing ccr1 should not be an issue.

Indanz: Yeah, if you remember the vale you wrote to ccr1 and the increment, you don't need to read the register and the drift problem goes away.

libplatsupport/src/plat/stm32mp2/ltimer.c
Indanz: Of course. But you don't need to read it twice.

bruelc: > Of course. But you don't need to read it twice.

Ah yes. I was thinking the compiler would optimize but it's a volatile. Sorry yes it can be optimized

libplatsupport/src/plat/stm32mp2/timer.c
bruelc: yeah. checking pending overflow seems OK

libplatsupport/src/plat/stm32mp2/timer.c
```diff
@@ -0,0 +1,102 @@
+/*
+ * Copyright 2026, STMicroelectronics
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+/* Use 32 bit TIM2 general timer to handle 64 bit timestamp with ARR
+   overflow in up-counting mode. CNT is reset at earch overflow.
+   32 bit timeout interrupt is supported with Channel 1 output compare. */
```
Indanz: I remember reading that there was a separate overflow/count reset interrupt, so you shouldn't need to use a separate channel just for this.

libplatsupport/src/plat/stm32mp2/timer.c
```diff
@@ -0,0 +1,102 @@
...
+    void *rcc = NULL;
+    ps_io_ops_t *io_ops = &ops;
+
+    /* TIM2 RCC */
+    MAP_IF_NULL(io_ops, RCC_TIM2, rcc);
+    if (rcc == NULL) {
+        return EINVAL;
+    }
+    RCC_ON(rcc);
+    asm volatile("isb" ::: "memory");
```
Indanz: I would add a comment why an `isb` is needed. Considering `RCC_ON` does a memory write, I would expect a `dsb`. An `isb` doesn't guarantee that the write finishes, only that the instruction finished. Things are a bit muddy for uncached memory writes, so in practice probably either works, but `dsb` seems more correct.

Indanz (approved):

libplatsupport/src/plat/stm32mp2/timer.c
bruelc: ok. This is to make sure that the interconnect rcc write is visible before access the tim. just checked, tf-a does a read-back, that should be enough

Indanz: A read-back is only guaranteed to be enough if the RCC is in the same peripheral as the TIM. If not, the CPU is allowed to reorder those memory accesses and do the TIM writes before the read-back finished.

Only way to make sure that the read-back works is by putting it in a while loop till it changes some value: The CPU is not allowed to do speculative accesses to device memory, so that would stop the CPU from issuing the TIM writes too early (because of the control dependency).

bruelc: Latest implementation uses a while a loop indeed. Although a single read in practice could be enough for the clock to propagate.

Indanz: Yes, but if you do a single read, the CPU is allowed to do the later TIM writes while it is waiting for the RCC read to finish. Which would have the same effect as having no read at all. Timing wise a single read would be enough, yes, but enforcing such timing is tricky on out-of-order CPUs.

bruelc:
Hum maybe a single read back and a memory barrier would be equivalent . I
used the loop for a different reason, because I de-assert  and enable the
clock in the same write to cover this additional dependency. The loop
should also cover the order problem you mention. Thank you for the
explanation

Le lun. 9 mars 2026 à 12:48, Indan Zupancic ***@***.***> a
écrit :

bruelc:
In the ref man 21.4.11 , the recommended sequence is to use a single dummy read after enabling the clock

PR #206, opened 2026-05-02
omap3+apq8064/mux.c: fix type signature

lsf37:
mux.h declares the first parameter of `feature_enable` as const, and `omap3_mux_feature_enable`, which is assigned to `feature_enable`, does not match. gcc-14 complains about the mismatch.

(same adjustment for apq8064_mux_feature_enable)


Indanz (approved):
Code doesn't make sense, but this is the simplest fix.

PR #207, opened 2026-09-22
Fix STM32MP2 timer cleanup.

bruelc:
Need to unmap the RCC vaddr when destroying the timer, to avoid a remap failure when trying to re-create it.
Also fix error return paths.


libplatsupport/src/plat/stm32mp2/timer.c
```diff
@@ -89,26 +89,38 @@ void enable_rcc(void *rcc)
...
+void stm32mp2_timer_destroy(stm32_t *stm32_timer,  ps_io_ops_t ops)
+{
+    if (stm32_timer->rcc == NULL) {
+        ZF_LOGE("Trying to destroy uninitialized timer");
+        return;
+    }
+
+    stm32_stop_timer(stm32_timer);
+
+    ps_io_unmap(&ops.io_mapper, stm32_timer->rcc, RCC_TIM2_SIZE);
```
Indanz:
```suggestion
    if (stm32_timer->rcc == NULL) {
        return;
    }
    stm32_stop_timer(stm32_timer);
    ps_io_unmap(&ops.io_mapper, stm32_timer->rcc, RCC_TIM2_SIZE);
```
But that's a matter of taste.

Indanz (approved):

libplatsupport/src/plat/stm32mp2/ltimer.c
```diff
@@ -171,14 +171,17 @@ int ltimer_default_init(ltimer_t *ltimer, ps_io_ops_t ops, ltimer_callback_fn_t
...
+        ps_free(&ops.malloc_ops, sizeof(stm32_ltimer_t), stm32_ltimer);
         return EIO;
     }
 
     rc = stm32mp2_timer_init(&stm32_ltimer->stm32_timer, ops);
     if (rc) {
         ZF_LOGE("Failed to init stm32 timeout timer");
-        destroy(&stm32_ltimer);
+        ps_irq_unregister(&ops.irq_ops, stm32_ltimer->irq_id);
+        ps_pmem_unmap(&ops, pmems, (void *) stm32_ltimer->stm32_timer.hw);
```
midnightveil: nit: these error cases should go through the timer teardown path, which should be modified to deal with partially initialised state. then you have one error path and no duplication of cleanup in multiple places.

this seems like what the destroy function was originally trying to do (?)

midnightveil (approved):

libplatsupport/src/plat/stm32mp2/ltimer.c
Indanz: Yes, but it has a `void*` as parameter and it was silently passed the wrong pointer before. This makes things explicit.

(I almost wrote the same comment myself.)

bruelc: > nit: these error cases should go through the timer teardown path, which should be modified to deal with partially initialised state. then you have one error path and no duplication of cleanup in multiple places.

do you mean with gotos  ? I was not sure, I'll update then

> 
> this seems like what the destroy function was originally trying to do (?)

yes, but now destroy() also unmap  the rcc, so we need to deal partially initialized paths

